PR #5, opened 2019-05-17
minimal stm32l0x2 support. Alternate Function setup for SPI pins

lthiery:
I've added in minimal support for the stm32l0x2. In particular, I've tested UART and SPI and done some pin-to-AF mappings for the chip.

In an effort to get SPI working I added in the alternate function configuration of the pins to the driver, similar to how it was done in UART.

Also, it appears that embedded-hal is deprecating the pin interfaces that this project uses, so there's a flood of warnings about that. I may be able to try to resolve that but just wanted to give you a heads up.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,10 @@ license = "0BSD"
 name = "stm32l0xx-hal"
 readme = "README.md"
 repository = "https://github.com/stm32-rs/stm32l0xx-hal"
-version = "0.1.0"
+version = "0.1.1"
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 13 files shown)


lthiery:
Before working on EXTI/GPIO, I wanted to get examples building. I also realized I forgot to do SPI definitions for stm32l0x1 feature.

arkorobotics:
Thanks for the PR! The changes look good, but I don't have the time at the moment to verify functionality on the stm32l0x1. I'm going to merge the changes now and test them when I return from my work trip.

PR #7, opened 2019-06-05
Fix examples

lthiery:
I tried to make examples work on both stm32l0x1 and stm32l0x2. Please test and make sure I didn't break them on your platform.

And please consider this PR _after_ #6


arkorobotics:
I see two active issues with this PR. The first being that the Travis CI is failing to build and the second being that the flash and ram sizes are too large for the stm32l0x1. How have other HAL projects handled this issue? Is there a way to define the flash/ram size based on the mcu feature parameter?

hannobraun:
@arkorobotics Here's how `nrf52832-hal` does it: One `memory.x` file per supported configuration, and a build script that selects the correct one based on Cargo features.

Here: https://github.com/nrf-rs/nrf52-hal/tree/4c21ba5f0b5ef4012770ee76db6a9ae1884dc6cd/nrf52832-hal

The relevant files being `memory_xxAA.x`, `memory_xxAB.x`, and `build.rs`.

arkorobotics:
Excellent suggestion @hannobraun! I got it working in my local branch after discovering (aka getting stuck on) the fact that the number of bytes read from each memory file must match... I added white spaces to get it to work, but I'm not sure that's the best solution. Thoughts?

Build error when memory file sizes don't match:
```
error[E0308]: match arms have incompatible types
  --> build.rs:18:13
   |
10 |        let linker = match (cfg!(feature = "stm32l0x1"), cfg!(feature = "stm32l0x2")) {
   |   __________________-
11 |  |         (false, false) | (true, true) => {
   |  |__________________________________________-
12 | ||             panic!("\n\nMust select exactly one package for linker script generation!\nChoices: 'stm32l0x1' or 'stm32l0x2'\n\n");
13 | ||         }
   | ||_________- this is found to be of type `_`
14 |  |         (true, false) => {
15 |  |             include_bytes!("memory_l0x1.x")
   |  |             ------------------------------- this is found to be of type `_`
...   |
18 |  |             include_bytes!("memory_l0x2.x")
   |  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a fixed size of 95 elements, found one with 97 elements
19 |  |         }
20 |  |     };
   |  |_____- `match` arms have incompatible types
   |
   = note: expected type `&[u8; 95]`
              found type `&[u8; 97]`
```

hannobraun:
@arkorobotics Ha, that's funny, I didn't consider that :-)

Probably best to use [`std::fs::copy`](https://doc.rust-lang.org/std/fs/fn.copy.html) then.

hannobraun:
Different (maybe simpler) idea: Can you do `&include_bytes!(...)` or `includes_bytes!(...).as_ref()` to turn it into a slice?

arkorobotics:
> Different (maybe simpler) idea: Can you do `&include_bytes!(...)` or `includes_bytes!(...).as_ref()` to turn it into a slice?

`includes_bytes!(...).as_ref()` worked like a charm, thanks!

lthiery:
@hannobraun i should close this PR, right? I think that you already upstreamed similar stuff. I think I squeeze an EXTI bug fix in this too but I can track that down this week and upstream.

hannobraun:
@lthiery 
> i should close this PR, right?

I don't know. There's definitely some overlap with what I've done (which got merged), but there's also a lot of stuff in this PR that looks unfamiliar.

PR #11, opened 2019-07-03
Bug fixes to EXTI and GPIO

lthiery:
Sorry - I just realized with this PR that my `cargo fmt` touched a lot of stuff that I didn't mean to touch. That being said, it's probably worth adding a `cargo fmt` check to CI to I've added that to `.travis.yml`.

Other than that, my main goal was to rebase the branch that I've been building a project off of. Mainly bug fixes to EXTI and GPIO but I also make the UART peripheral names public.

One problem that is _not_ fixed in this branch is that the `flush` in `write_str` seems to throw errors consistently, at least on my platform.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -4,16 +4,13 @@ use core::ops::Deref;
 
 use crate::hal::blocking::i2c::{Read, Write, WriteRead};
 
-use cast::u8;
-use crate::gpio::gpioa::{PA9, PA10, PA13};
+use crate::gpio::gpioa::{PA10, PA13, PA9};
 use crate::gpio::gpiob::{PB6, PB7};
 use crate::gpio::{AltMode, OpenDrain, Output};
-use crate::pac::{
-    i2c1::RegisterBlock,
-    I2C1,
-};
+use crate::pac::{i2c1::RegisterBlock, I2C1};
 use crate::rcc::Rcc;
 use crate::time::Hertz;
+use cast::u8;
 
 #[cfg(feature = "stm32l0x1")]
 use crate::gpio::gpioa::PA4;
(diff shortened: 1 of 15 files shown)


hannobraun:
I'd like to note that I think the `rustfmt` changes, as well as requiring code to be `rustfmt`ed via the CI, are a really bad idea. It makes the code consistent at the cost of readability. I must be missing something really obvious because so many other developers don't question this, but I really don't understand what the point of consistency is, if it comes at the cost of readability.

src/i2c.rs
```diff
@@ -126,12 +109,16 @@ impl<I, SDA, SCL> I2c<I, SDA, SCL>
...
+            w.presc()
+                .bits(presc)
+                .scll()
+                .bits(scll)
+                .sclh()
+                .bits(sclh)
+                .sdadel()
+                .bits(sdadel)
+                .scldel()
+                .bits(scldel)
```
hannobraun: This is an example of code that has become much less readable as a result of running `rustfmt` (see my top-level comment in the pull request). Arguably there are other kinds of examples too, but it's really obvious and egregious with these svd2rust-generated APIs.

arkorobotics: Great point, I had not considered this and I'm inclined to agree that readability is far more important than consistency. I'll start following the readable format moving forward.

lthiery:
The only reason that I put `cargo fmt` into CI is that it short-circuits me from spending my own time formatting code and keeps time from being spent debating formatting. 

But since we are debating automated formatting, I may as well throw in my 2 cents: I believe consistent code leads to readability. @hannobraun 's example is a good counter-example, _but_ most the time, it does just fine and in these special cases, I've learned to deal with it as I believe the benefits outweigh the costs.

Anyway, I'm content removing it from CI and I can just use it to clean up my own code.

hannobraun:
I think that's actually a good point. If automated formatting is a net-positive at zero effort, it might actually be better overall than more readable formatting at higher effort.

I guess where to draw the line here, i.e. how much reduced readability to accept for the benefit of not having to format manually, comes down to personal preference. I would argue, however, that in any codebase dominated by the use of svd2rust APIs (like a HAL), the damage is greater than in most other cases.

As for me personally, I don't mind doing the formatting myself, as I find it comes to me easily and I rarely think about it. When I do have to think about it, it's usually a case where the readability benefits from a little effort, and that is worth it to me.

In any case, I don't mean to force my formatting choices on anyone. I do strongly believe that requiring `cargo fmt` in CI is not the right choice for this project, and that running `cargo fmt` on a HAL is most often a net-negative.

lthiery:
>  I would argue, however, that in any codebase dominated by the use of svd2rust APIs (like a HAL), the damage is greater than in most other cases.

I find that to be a very reasonable assessment.

PR #12, opened 2019-07-05
Add basic AES API

hannobraun:
Add a basic AES API. So far, only CTR mode is supported.


hannobraun:
I also have some additional commits here, for testing this change on the Tlera Corp Gnat:
https://github.com/braun-embedded/stm32l0xx-hal/compare/aes...braun-embedded:aes-gnat

cc @lthiery

PR #19, opened 2019-07-23
Make DMA-based AES API more ergonomic

hannobraun:
This pull request changes the AES API to work with `u8` buffers, instead of `u32` buffers, which should make it easier to use in most scenarios.

I've also snuck in a few commits that fix and improve some other things related to AES and DMA. It didn't seem worth it to create a separate pull request for those, especially since I've already got a bit a bit of a traffic jam going :-)

This pull request is based on #16. I recommend reviewing/merging that one first.

cc @lthiery

PR #16, opened 2019-07-16
Add support for interrupts to DMA API

hannobraun:
This pull request is based on #14. I recommend reviewing/merging that pull request first, before this one.

I have additional commits to test this against the Tlera Corp Gnat here:
https://github.com/braun-embedded/stm32l0xx-hal/compare/dma-interrupts...braun-embedded:dma-interrupts-gnat

cc @lthiery

PR #20, opened 2019-07-24
Add DMA support to USART API

hannobraun:
Also contains some small cleanups of the `serial` module.

This pull request is based on #19. I recommend reviewing/merging that pull request first.

cc @lthiery


hannobraun:
I just rebased this on the latest `master`. It still contains all the commits from #16 and #19. Normally I mark pull requests that are based on others as a draft, but I forgot to do so in this case.

PR #21, opened 2019-07-29
Clean up PWM code, fix race conditions, improve STM32L0x2 support

hannobraun:
This is a big clean-up effort of the `pwm` module. It fixes the following problems:
- The module was a bit of a mess, with lots of duplicated code.
- Some of the `unsafe` access to timer registers was unsound, as it didn't take race conditions into account.
- The module was not flexible enough to support STM32L0x2 fully.

Overall, I believe that this new approach, with each channel being its own struct that can be configured separately, is a step in the right direction. In fact, I think we should consider using the same approach for the `timer` module, maybe even merge both modules. (I'm not sure if the hardware supports this, but if both modules were merged it could be possible to use one channel for PWM, another for something else, e.g. as a hardware trigger for ADC.)

cc @lthiery

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,7 +109,3 @@ required-features = ["rt"]
 [[example]]
 name = "adc_pwm"
 required-features = ["stm32l0x1"]
-
-[[example]]
-name = "pwm"
-required-features = ["stm32l0x1"]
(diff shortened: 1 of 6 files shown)


hannobraun:
Rebased on latest master.

PR #22, opened 2019-07-31
Add support for all remaining AES modes

hannobraun:
This pull request is based on #19. I recommend reviewing/merging that one first, before taking a look at this one.

cc @lthiery


arkorobotics:
Thanks so much for adding the AES support and for cleaning up the PWM source code and examples!

hannobraun:
> Thanks so much for adding the AES support and for cleaning up the PWM source code and examples!

No problem. Thank my client, they paid for it :-)

Thank you for creating `stm32l0xx-hal` in the first place, and for keeping up with all my pull requests!

PR #25, opened 2019-08-07
Add PWR API, support for wake-up via EXTI

hannobraun:
Add a PRW API with support for Sleep and Stop modes, as well as a method that goes to sleep and waits for an EXTI interrupt. I also snuck in some clean-up of the `prelude` module, while I was in there.

cc @lthiery

### DIFF
--- a/examples/exti_wakeup.rs
+++ b/examples/exti_wakeup.rs
@@ -62,13 +62,15 @@ fn main() -> ! {
         interrupt::free(|_| {
             nvic.enable(Interrupt::EXTI2_3);
 
-            pwr.enter_stop_mode(
-                &mut scb,
-                &mut rcc,
-                pwr::StopModeConfig {
-                    ultra_low_power: true,
-                },
-            );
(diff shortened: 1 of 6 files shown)


hannobraun:
I just noticed I forgot to enable the PWR peripheral clock. I did not notice earlier because everything worked as expected. This could happen because the main work here is being done by SCB, RCC, and `wfi`, and the PWR interface only serves a secondary role.

Only Stop mode is affected. I _think_ that Stop mode is actually still entered (because I explicitly saw a slower clock after wake-up before I figured out I need to set `STOPWUCK`), but the extra configuration is probably non-functional. I'll push another commit to fix this shortly.

hannobraun:
Pushed another commit. As best as I can tell, everything should work as expected. Once I've implemented more power modes (should happen in a few weeks), I'll do more testing with some actual current measurement.

arkorobotics:
Nice! Thanks so much for the PR.

PR #30, opened 2019-08-26
Fix NVIC deprecation warnings

Disasm:
This is a breaking change, because EXTI API is changed.

### DIFF
--- a/examples/aes_ctr_dma.rs
+++ b/examples/aes_ctr_dma.rs
@@ -17,6 +17,7 @@ use aligned::{
 use cortex_m::{
     asm,
     interrupt,
+    peripheral::NVIC,
 };
 use cortex_m_rt::entry;
 use stm32l0xx_hal::{
@@ -39,7 +40,6 @@ use stm32l0xx_hal::{
 
 #[entry]
 fn main() -> ! {
-    let mut cp = pac::CorePeripherals::take().unwrap();
     let     dp = pac::Peripherals::take().unwrap();
 
     let mut rcc = dp.RCC.freeze(Config::hsi16());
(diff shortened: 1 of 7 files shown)


Disasm:
@arkorobotics could you review this? I'd like to bump `stm32-usbd` dependencies and `stm32l0xx-hal
` is the only hal I wait for.

arkorobotics:
The version bump to 0.3.0 is a large leap. If anything, we'd bump it from 0.2.0 to 0.2.1.

As for the NVIC changes, they look good to me at a quick glance, but I would like @hannobraun to review it before pulling.

hannobraun:
I've taken a short look and can't see any issues. Seems fairly non-controversial, as it just replaces deprecated functions with their replacements. The only thing I'd look into is why `NVIC::unmask` is unsafe, and how that matches our usage.

I don't have time to take a closer look right now. I'm on vacation and will be back in the office on September 9.

> The version bump to 0.3.0 is a large leap. If anything, we'd bump it from 0.2.0 to 0.2.1.

I'm pretty sure I've made breaking changes since the last release. This would require a bump to 0.3.0 per semantic versioning.

Disasm:
At least `stm32l0` dependency update is a breaking change: https://github.com/stm32-rs/stm32l0xx-hal/commit/e8a6522fb03749c550fbd4a38ac81be76936cca9

arkorobotics:
Copy that. I'll merge this and bump to version 0.3.0.

arkorobotics:
@Disasm This merge has been rolled into 0.3.0. Please let me know if there are any issues https://crates.io/crates/stm32l0xx-hal

PR #31, opened 2019-08-26
dma fix and dma_async example

lthiery:
I was trying to put together an 'async' DMA example that basically continues looping without waiting for DMAs to finish but I ran into some issues.

As according to the comments for `wait`:
```
    /// [wait] will busily wait until the transfer is finished. If you
    /// don't want this, please call this function only once you know that the
    /// transfer has finished.
```

I assumed the only other Transfer function, `is_active`, was the intended way of checking this (@hannobraun). Unfortunately, it modifies the registers when returning false, meaning a non-blocking call to `wait` after checking `active` would get stuck in `wait`. That is to say, the following fails:
```
if !transfer.is_active() {
   // gets stuck because internal calls to `is_active` always returns true
   transfer.wait();
}
```

I've modified `is_active` to have an flag for clearing the registers, making usage within `wait` clear flags, but calls to Transfer::is_active will not clear flags.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,9 @@ version = "0.8.0"
 
 [dev-dependencies]
 aligned = "0.3.1"
+heapless = "0.5.0"
 panic-halt = "0.2.0"
 panic-semihosting = "0.5.1"
-
 
 [features]
 # Miscellaneaous features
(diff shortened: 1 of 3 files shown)


hannobraun:
Yes, `is_active` was intended to do exactly that. And it did at one point, but I must have broken it with later changes.

I haven't reviewed the example, but the changes to the DMA API look good to me.

arkorobotics:
CI appears to be failing for the L0x1, can we resolve this before merging?

Issue #34, opened 2019-09-02
docs.rs failed to build stm32l0xx-hal-0.3.0


tkeksa:
See https://docs.rs/crate/stm32l0xx-hal/0.3.0/builds/182532
The problem seems to be custom build.rs which requires exactly one feature enabled but there is:
```
[package.metadata.docs.rs]
features = ["stm32l0x1", "stm32l0x2", "stm32l0x3", "rt"]
```
Probably either just one feature should be enabled for docs build (as it has e.g. stm32f1xx_hal) or custom build.rs should be disabled for docs somehow.

arkorobotics:
Thanks for the ticket! This issue has been resolved. Please see docs.rs for version [0.3.1](https://docs.rs/crate/stm32l0xx-hal/0.3.1/).

PR #37, opened 2019-09-09
Try fixing the docs.rs build

jonas-schievink:
Only enable the `stm32l0x3` feature.

cc https://github.com/stm32-rs/stm32l0xx-hal/issues/34

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ repository = "https://github.com/stm32-rs/stm32l0xx-hal"
 version = "0.3.0"
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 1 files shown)


arkorobotics:
Not to step on your PR, but I would like to enable the `stm32l0x1` feature instead (see commit [eaeac86](https://github.com/stm32-rs/stm32l0xx-hal/commit/eaeac86139ff468380880ad3e58768c6368df04c)) as the `stm32l0x3` feature is fairly new and I haven't had a chance to review which feature has the most coverage.

jonas-schievink:
Sure, that makes sense too

PR #38, opened 2019-09-09
Make RTFM a dev-dependency

jonas-schievink:
It isn't needed as a normal dependency, and pulls in an old cortex-m version (0.5.x).


arkorobotics:
Great point, thanks!

PR #44, opened 2019-09-30
RTC bugfix and improvements

hannobraun:


hannobraun:
Oops, forgot something: cc @lthiery

PR #45, opened 2019-10-01
Add support for low-power run/sleep mode

hannobraun:
This adds complete support for entering all low-power modes supported by the hardware. Also contains a commit with a small PWR-related fix.


hannobraun:
cc @lthiery

arkorobotics:
Thanks!

PR #47, opened 2019-10-15
Add SPI2, LPUART, and I2C definitions for stm32l0x2/3

lthiery:
Tested only on stm32l0x2, but as I understand it, stm32l0x3 is a superset of stm32l0x2?


lthiery:
I ran into a few more pins I needed for a board I was bringing up. All have been tested on stm32lx02 only but I am assuming stm32l0x3 is a superset.

PR #48, opened 2019-10-18
Add Cargo feature to disable linker script

hannobraun:
This is useful if the user wants to supply their own linker script.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -25,24 +25,27 @@ fn main() {
         panic!("\n\nMust select exactly one package for linker script generation!\nChoices: 'stm32l0x1' or 'stm32l0x2' or 'stm32l0x3'\n\n");
     }
 
-    let linker = if cfg!(feature = "stm32l0x1") {
-        include_bytes!("memory_l0x1.x").as_ref()
-    } else if cfg!(feature = "stm32l0x2") {
-        include_bytes!("memory_l0x2.x").as_ref()
-    } else if cfg!(feature = "stm32l0x3") {
-        include_bytes!("memory_l0x3.x").as_ref()
-    } else {
-        unreachable!();
-    };
+    if !cfg!(feature = "disable-linker-script") {
+        let linker = if cfg!(feature = "stm32l0x1") {
+            include_bytes!("memory_l0x1.x").as_ref()
+        } else if cfg!(feature = "stm32l0x2") {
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -25,24 +25,26 @@ fn main() {
...
+            unreachable!();
+        };
+
+        File::create(out.join("memory.x"))
+            .unwrap()
+            .write_all(linker)
+            .unwrap();
+        println!("cargo:rustc-link-search={}", out.display());
+
+        println!("cargo:rerun-if-changed=build.rs");
```
jonas-schievink: It might be better to always output this line, since otherwise the build script will be rerun on every build I think.

hannobraun: Good catch, but actually this line is redundant, per https://doc.rust-lang.org/cargo/reference/build-scripts.html. I've pushed another commit that removes it entirely.

jonas-schievink: Hmm, that says:

> Normally build scripts are re-run if any file inside the crate root changes

And the `cargo:rerun-if-changed=build.rs` note says:

> (unless you want to ignore changes in all other files except for build.rs)

So it's not quite redundant. It probably doesn't hurt to run the build script every time any file changes, but it's also not necessary.

hannobraun: My interpretation of the `rerun-if-changed` documentation is this:
- By default, the build script is re-run on every change to any file.
- You can restrict this, by issuing `rerun-if-changed`. Then reruns are limited to changes to the specified files.
- **Except** that changes to `build.rs` will always trigger a re-run.
- Therefore `cargo:rerun-if-changed=build.rs` is redundant, unless it is the only `rerun-if-changed` directory issued.

arkorobotics:
Ready to pull?

hannobraun:
Not yet. I think I might have misunderstood what @jonas-schievink was saying. I'll look into it later.

build.rs
hannobraun: Okay, I think I completely misunderstood what you were saying. I was thinking about what happened inside the `if`, and yes, the `cargo:rerun-if-changed=build.rs` is redundant there. It is not redundant if `disable-linker-script` is enabled, as then the linker script is re-run on every file change.

I'm going to push a new commit soon.

hannobraun:
I pushed a new version with the fix @jonas-schievink suggested and squashed the commits. Should be ready to merge now.

PR #50, opened 2019-10-28
RNG driver

lthiery:
Provides a simple synchronous RNG driver and example. Refactor enable 48 MHz code from USB driver to make it accessible via RCC.

### DIFF
--- a/examples/rng.rs
+++ b/examples/rng.rs
@@ -1,0 +1,37 @@
+#![no_main]
+#![no_std]
+//#![deny(warnings)]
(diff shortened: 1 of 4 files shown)


hannobraun:
Travis is failing. Otherwise this looks good.

hannobraun:
There's a problem with this PR that I didn't see before: Both the RNG and USB modules consume `pac::CRS`, which means you can't use them both in the same program. I'll look into it.

lthiery:
Closed in favor of #53

PR #51, opened 2019-11-01
Update to stm32l0 0.9.0

jonas-schievink:


jonas-schievink:
Okay, now that stm32l0 0.9.0 is published this should be ready for review @arkorobotics

arkorobotics:
Thanks for taking the time to update the hal to use stm32l0 0.9.0! Looks good to me.

PR #55, opened 2019-11-05
Mark a few functions as inline

jonas-schievink:
This saves 1.5 KiB of binary size in the `pwr.rs` example.

See comments for explanations.


arkorobotics:
That's a significant binary size saving, thanks!

PR #56, opened 2019-11-13
Add USB driver and example

Disasm:
See also: https://github.com/stm32-rs/stm32-usbd/pull/16

cc @hannobraun

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ repository = "https://github.com/stm32-rs/stm32l0xx-hal"
 version = "0.4.0"
 
 [package.metadata.docs.rs]
-features = ["stm32l0x1", "rt"]
+features = ["stm32l0x2", "rt", "stm32-usbd"]
 
 [dependencies]
 as-slice = "0.1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,8 @@ heapless = "0.5.0"
 panic-halt = "0.2.0"
 panic-semihosting = "0.5.1"
 cortex-m-rtfm = "0.4.3"
+usb-device = "0.2.3"
+usbd-serial = "0.1.0"
 
 [features]
 # Miscellaneaous features
@@ -142,3 +144,7 @@ required-features = ["rt"]
 [[example]]
 name = "adc_pwm"
 required-features = ["stm32l0x1"]
+
+[[example]]
+name = "usb_serial"
+required-features = ["rt", "stm32l0x2", "stm32-usbd"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ cortex-m-semihosting = "0.3.2"
 void = { version = "1.0.2", default-features = false }
 cast = { version = "0.2.2", default-features = false }
 nb = "0.1.2"
+stm32-usbd = { version = "0.5.0", features = ["ram_access_2x16"], optional = true }
 
 [dependencies.stm32l0]
 version = "0.8.0"
--- a/examples/usb_serial.rs
+++ b/examples/usb_serial.rs
@@ -6,7 +6,7 @@ extern crate panic_semihosting;
 
 use cortex_m_rt::entry;
 use stm32l0xx_hal::usb::{Peripheral, UsbBus};
-use stm32l0xx_hal::{pac, prelude::*, rcc};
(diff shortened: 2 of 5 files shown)


Disasm:
I haven't checked this USB example (I have no board), it was ported from stm32f0xx-hal using STM32L0-specific initialization code from [Hanno's test_class example](https://github.com/Disasm/stm32-usbd-tests/blob/master/stm32l0x2/src/main.rs).

hannobraun:
This is fantastic! https://github.com/stm32-rs/stm32-usbd/pull/16 is a great and necessary change. I also started experimenting with this a while ago, but didn't finish a pull request. Glad to see we were thinking along the same lines.

I've just glanced over the code so far. I'll do a proper review and test this on hardware soon.

(And sorry for the late reply. I was at RustFest and am now slowly working through the backlog.)

src/usb.rs
```diff
@@ -1,39 +1,54 @@
...
+            rcc.apb1enr.modify(|_, w| w.usben().set_bit());
+
+            // Reset USB peripheral
+            rcc.apb1rstr.modify(|_, w| w.usbrst().set_bit());
+            rcc.apb1rstr.modify(|_, w| w.usbrst().clear_bit());
+        });
+    }
+
+    fn startup_delay() {
+        // There is a chip specific startup delay. For STM32F103xx it's 1µs and this should wait for
```
hannobraun:
```suggestion
        // There is a chip specific startup delay. For STM32L0x2/x3 it's 1µs and this should wait for
```

This seems to be true for all STM32L0x2/x3, according to their data sheets.

src/usb.rs
```diff
@@ -1,39 +1,54 @@
...
+        // at least that long.
+        cortex_m::asm::delay(72);
```
hannobraun:
```suggestion
        cortex_m::asm::delay(32);
```

All L0's run at up to 32 MHz.

examples/usb_serial.rs
```diff
@@ -0,0 +1,66 @@
+//! CDC-ACM serial port example using polling in a busy loop.
+#![no_std]
+#![no_main]
+
+extern crate panic_semihosting;
+
+use cortex_m_rt::entry;
+use stm32l0xx_hal::usb::{Peripheral, UsbBus};
+use stm32l0xx_hal::{pac, prelude::*, rcc};
```
hannobraun:
```suggestion
use stm32l0xx_hal::{pac, prelude::*, rcc, syscfg::SYSCFG};
```
Required below. See other comment.

examples/usb_serial.rs
```diff
@@ -0,0 +1,66 @@
...
+use usb_device::prelude::*;
+use usbd_serial::{SerialPort, USB_CLASS_CDC};
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    let mut rcc = dp.RCC.freeze(rcc::Config::hsi16());
```
hannobraun: The example doesn't set up the HSI48 clock:
```suggestion
    let mut rcc = dp.RCC.freeze(rcc::Config::hsi16());
    let mut syscfg = SYSCFG::new(dp.SYSCFG_COMP, &mut rcc);
    rcc.enable_hsi48(&mut syscfg, dp.CRS);
```

The old API enforced this by requiring a `HSI48` as an argument. I think it's unfortunate that the new API doesn't do this, but I think this can be improved upon later.

Cargo.toml
```diff
@@ -35,6 +35,7 @@ cortex-m-semihosting = "0.3.2"
 void = { version = "1.0.2", default-features = false }
 cast = { version = "0.2.2", default-features = false }
 nb = "0.1.2"
+stm32-usbd = { version = "0.5.0", features = ["ram_access_2x16"], optional = true }
```
hannobraun: Not sure if the dependency needs to be optional. Yes, it's a heavy one, but that's only relevant on the first compile, as embedded applications will use LTO to remove unused stuff anyway. Maybe I'm wrong.

Disasm: It's just a dependency tree (==build time) optimization, it doesn't affect the code size.

Disasm:
Thanks for the suggestions! Applied them all, could you test one more time?

arkorobotics:
Awesome work! Thanks everyone!

PR #57, opened 2019-11-18
Fix names of GPIOD types/fields

jonas-schievink:
These were referring to Port C, not good


jonas-schievink:
Let me clean up the whole gpio module while I'm at it...

PR #58, opened 2019-11-18
GPIO cleanup and fixes

jonas-schievink:


src/gpio.rs
```diff
@@ -79,10 +79,11 @@ pub enum Port {
...
 }
 
+/// Error type used by the GPIO trait impls.
+///
+/// GPIO operations cannot fail, so this is uninhabited.
 #[derive(Debug)]
-pub enum Error {
-    Foo,
-}
+pub enum Error {}
```
hannobraun: Why not `void::Void`? We already depend on it, and there's stuff like `void_unwrap` in the crate, which seems neat.

Issue #42, opened 2019-09-24
EXTI API is partially broken, could use an update


hannobraun:
I've been planning to update the EXTI API for a while now, as I believe that it has started to outgrow its initial design. I started working on that, but it turns out I bit off more than I can chew.

I'm pausing this work until I can come up with some new ideas. In the meantime, I'm opening this issue to document the current shortcomings of the API, and the problems I had coming up with a better design.

Here's what's wrong with EXTI, as I see it:
- The API is partially broken. Some of its methods rely on bit-banding, but everything I read says that bit-banding is only supported on Cortex-M3/M4. I couldn't find any evidence that STM32L0 is an exception.
- The API is focused on configurable interrupt lines, more specifically GPIO interrupts. Direct interrupt lines aren't really considered. This means that a user has to pass all kinds of arguments that aren't actually needed.

In essence, there are three kind of interrupt lines:
- Direct lines, which can only be enabled/disabled.
- Configurable lines, which can be configured to trigger on rising and/or falling edges.
- GPIO lines, which are configurable lines that also need special handling in SYSCFG, as all GPIO pins are multiplexed onto 16 EXTI lines.

The challenge I had was coming up with a design that:
1. Doesn't require the user to pass arguments that are not actually used for a given interrupt line. That isn't just inconvenient but also error-prone, as it can easily lead to misunderstandings.
2. Doesn't duplicate code all over the place.
3. Does 1. and 2. without becoming super-complicated.

I'll think about this in the back of my mind for a while before I give this another go. If anyone has any ideas in the meantime, feel free to comment.

jonas-schievink:
(working on this now)

Issue #52, opened 2019-11-01
Timers can't sleep for >1 second


jonas-schievink:
The hal::timer::CountDown impl of the timers uses Hertz as the Time parameter, which is a newtype around u32. This makes the longest timeout duration 1 Hz or 1 second. It would be useful to support this.

romixlab:
Same problem with watchdog

PR #59, opened 2019-11-21
Improve USB module

hannobraun:
This improves the `usb` module, bringing it more in line with other modules and making the API less error-prone.

### DIFF
--- a/examples/usb_serial.rs
+++ b/examples/usb_serial.rs
@@ -5,7 +5,7 @@
 extern crate panic_semihosting;
 
 use cortex_m_rt::entry;
-use stm32l0xx_hal::usb::{Peripheral, UsbBus};
+use stm32l0xx_hal::usb::{USB, UsbBus};
 use stm32l0xx_hal::{pac, prelude::*, rcc, syscfg::SYSCFG};
 use usb_device::prelude::*;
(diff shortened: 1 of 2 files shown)


arkorobotics:
Great! Given all the recent additions, I think it's time to release the current repo (after merging this) as v0.5.0. Thoughts? Do we want to close out all open issues before doing so?

hannobraun:
> Great! Given all the recent additions, I think it's time to release the current repo (after merging this) as v0.5.0. Thoughts? Do we want to close out all open issues before doing so?

Sorry for the late reply! A new release sounds good to me.

Regarding the open issues, I definitely plan to close #42 before the year is over, but I wouldn't block the release on that. I currently have no plans to work on #52.

I plan to add a Flash API very soon (pull request incoming this week, if nothing goes wrong, but who knows). Don't know, if you want to wait for that.

Disasm:
Is HSI48 absolutely required for USB to work on L0?

hannobraun:
> Is HSI48 absolutely required for USB to work on L0?

I think there are other clocks that can be used in its place, although I don't recall the details right now. HSI48 is the only one currently supported in the HAL API.

arkorobotics:
@hannobraun No worries. Let's wait for the Flash API. That sounds like a great addition before the next rev (no pressure 😛).

PR #60, opened 2019-11-27
Add Flash API

hannobraun:
Please note that implementing one of the features, writing half-pages to Flash, required some unconventional thinking, i.e. C source code and a binary blob. Please see the included comments for an explanation of why I found it necessary.


hannobraun:
Pushed two new commits: I forgot a critical section and found a typo.

arkorobotics:
Great work! Sounds like this will pave the way for folks to write bootloaders :) Writing half-pages to flash using only Rust sounds like an interesting problem. Hope we can patch the C code in the future!

PR #64, opened 2019-12-06
Rewrite the EXTI interface

### LINKED ISSUE #42: EXTI API is partially broken, could use an update

jonas-schievink:
This essentially does a few things:
* Stops using `u8` as the type for interrupt lines, replacing it with 3 types implementing a common trait. These 3 types reflect @hannobraun's "three kinds of interrupt lines" mentioned in #42. This allows getting rid of the `assert_line_valid` function and removes panic branches, made possible by the more type-safe API.
* Gets rid of the bit-banding interface, using `.modify` instead.
* Changes the EXTI API to be sound(-er) by taking `&mut self` when required
* Loosens other parts of the API (namely, clearing and pending interrupts) to not take `self` at all, mirroring the NVIC API provided by `cortex-m`
* Replaces the "get mask of pending interrupts" function with `is_pending`, again mirroring the NVIC API

Closes #42


lthiery:
Much better! Nice

hannobraun:
I agree, looks much better!

arkorobotics:
Great work! Looks awesome. 

I pulled the two previous PRs and it seems there's a minor merge conflict:
```
<<<<<<< exti
pub use crate::pac as device;
pub use crate::pac as stm32;
=======
mod bb;
>>>>>>> master
```

PR #65, opened 2019-12-11
Fix PLL enable.

craigjb:
This change fixes a bug wherein the PLL is never ready since the code to enable the PLL has some problematic writes that turn off the HSE or HSI osc input.

The PLL enable code included `self.cr.write` where we needed `self.cr.modify` to prevent the other register bits from resetting to default. The svd2rust register write resets everything. Modify changes bits without touching others.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -2,7 +2,10 @@ use crate::pac::RCC;
 use crate::time::{Hertz, U32Ext};
 
 #[cfg(any(feature = "stm32l0x2", feature = "stm32l0x3"))]
-use crate::{pac::CRS, syscfg::SYSCFG};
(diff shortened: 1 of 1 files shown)


arkorobotics:
Good catch, thanks for the fix! 

In the interest of formatting and consistency, I have one request: 

Please use:
```
use crate::{
    pac::CRS,
    syscfg::SYSCFG,
};
```

Instead of:
```
 use crate::{pac::CRS, syscfg::SYSCFG};
```

craigjb:
No problem. I fixed the formatting. Rustfmt was auto-formatting that away

arkorobotics:
Thanks!

PR #66, opened 2020-01-06
gpio: Enable all GPIO for all targets

dbrgn:
For some reason, GPIOC and GPIOD were disabled for the stm32l0x1, but these GPIO banks are present in the PAC (and in the datasheet).

### DIFF
--- a/examples/serial_dma_async.rs
+++ b/examples/serial_dma_async.rs
@@ -62,8 +62,8 @@ fn main() -> ! {
     // enqueue as many buffers as available into rx_buffers
     unsafe {
         // putting the same pointer in here twice would be a big mistake
-        rx_buffers.enqueue(Pin::new(&mut BUFFER_1)).unwrap();;
-        rx_buffers.enqueue(Pin::new(&mut BUFFER_2)).unwrap();;
+        rx_buffers.enqueue(Pin::new(&mut BUFFER_1)).unwrap();
+        rx_buffers.enqueue(Pin::new(&mut BUFFER_2)).unwrap();
     }
 
     let dma_handle = &mut dma.handle;
--- a/src/exti.rs
+++ b/src/exti.rs
@@ -57,13 +57,9 @@ impl ExtiExt for EXTI {
         let port_bm = match port {
             gpio::Port::PA => 0,
             gpio::Port::PB => 1,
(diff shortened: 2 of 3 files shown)


dbrgn:
Seems I broke something on the stm32l0x2. I'll check it out tomorrow.

jonas-schievink:
It looks like even GPIOE and GPIOH are present on the bigger L0x1 chips, so maybe all the `#[cfg]`s regarding that should be removed?

dbrgn:
Ah, you're right, I missed that!

dbrgn:
> Seems I broke something on the stm32l0x2. I'll check it out tomorrow.

It wasn't in my code, but I fixed it anyways.

> It looks like even GPIOE and GPIOH are present on the bigger L0x1 chips

Verified against the reference manuals and fixed.

Do you plan to limit the availability of GPIO pins depending on the package in the future?

hannobraun:
> Do you plan to limit the availability of GPIO pins depending on the package in the future?

I can't speak to anyone's plans, but I'd like to note there's precedent for going full bananas with packages: https://github.com/stm32-rs/stm32l0xx-hal/blob/68eeba6cd8c71a22b365d15dfe4ae9c4bb62888a/src/adc.rs#L223-#L299

jonas-schievink:
> It looks like even GPIOE and GPIOH are present on the bigger L0x1 chips, so maybe all the `#[cfg]`s regarding that should be removed?

This is technically true, but I think there is no released L0x1 chip that actually has those ports, so maybe it doesn't make sense to expose them?

dbrgn:
> This is technically true, but I think there is no released L0x1 chip that actually has those ports, so maybe it doesn't make sense to expose them?

Hm? There's the [STM32L071V8](https://www.st.com/content/st_com/en/products/microcontrollers-microprocessors/stm32-32-bit-arm-cortex-mcus/stm32-ultra-low-power-mcus/stm32l0-series/stm32l0x1/stm32l071v8.html) with an LQFP100 package.

image

jonas-schievink:
Oh, never mind then. I was confused by ST's product selector.

arkorobotics:
Good catch! Thanks for fixing the serial dma async bug as well. Not sure how that one slipped through the cracks.

PR #68, opened 2020-01-07
Serial: Replace `Pins` trait with `TxPin` / `RxPin`

### LINKED ISSUE #67: Serial `Pins` design discussion
Right now the `Pins` trait looks like this:

```rust
pub trait Pins<USART> {
    fn setup(&self);
}
```

It is then implemented for pin pairs using a macro:

```rust
macro_rules! impl_pins {
    ($($instance:ty, $tx:ident, $rx:ident, $alt:ident;)*) => {
        $(
            impl<Tx, Rx> Pins<$instance> for ($tx<Tx>, $rx<Rx>) {
                fn setup(&self) {
                    self.0.set_alt_mode(AltMode::$alt);
                    self.1.set_alt_mode(AltMode::$alt);
                }
            }
        )*
    }
}

impl_pins!(
    USART1,  PA9,  PA10, AF4;
    USART1,  PB6,  PB7,  AF0;
    USART2,  PA2,  PA3,  AF4;
    USART2,  PA14, PA15, AF4;
    USART2,  PD5,  PD6,  AF0;
    // ..
);
```

The problem with this is, that it assumes that a single pin pair can be used for an USART peripheral. However, pins can be mixed, so I could use USART1 with PA9 and PB7.

One approach would be to generate `Pins` implementations for all permutations (which can go into dozens or even hundreds of impls), but that's a workaround for a design problem in my opinion: Different pins can act as `Tx` and as `Rx` for an USART independently. (You could even use multiple `Tx` pins for the same USART if you wanted...)

My suggestion would be to drop the `Pins<USART>` trait, and to add `TxPin<USART>` and `RxPin<USART>` traits instead.

Here's a quick way to play around with this idea without having to adjust the serial trait:

```rust
pub trait TxPin<USART> {
    fn setup(&self);
}

pub trait RxPin<USART> {
    fn setup(&self);
}

impl<USART, TX, RX> Pins<USART> for (TX, RX) where TX: TxPin<USART>, RX: RxPin<USART> {
    fn setup(&self) {
        self.0.setup();
        self.1.setup();
    }
}

impl<Mode> TxPin<USART1> for PA9<Mode> { fn setup(&self) { self.set_alt_mode(AltMode::AF4); } }
impl<Mode> RxPin<USART1> for PA10<Mode> { fn setup(&self) { self.set_alt_mode(AltMode::AF4); } }
impl<Mode> TxPin<USART1> for PB6<Mode> { fn setup(&self) { self.set_alt_mode(AltMode::AF0); } }
impl<Mode> RxPin<USART1> for PB7<Mode> { fn setup(&self) { self.set_alt_mode(AltMode::AF0); } }
```

What do you think about this? I can create a PR with a concrete implementation if you want.

If dropping the `Pins` directly is not an option, we could also deprecate it and implement the Serial trait for both `Pins` and `RxPin` / `TxPin`.

cc @rnestler

dbrgn:
Based on #66, merge that first. Fixes #67.

The previous approach with the `Pins` trait assumed that an USART
peripheral can only be used with specific pin pairs. However, pins can
be mixed, so you could use USART1 with PA9 and PB7.

So instead of the `Pins` trait, there are now separate `TxPin` and
`RxPin` traits. They are implemented for all pins of all U(S)ART
peripherals of the stm32l0xx family.

Pin mappings verified against datasheets of stm32l071kb, stm32l072vz
and stm32l083vz.

### DIFF
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -45,7 +45,8 @@ fn main() -> ! {
 
     let serial = dp.USART2
         .usart(
-            (gpioa.pa2, gpioa.pa3),
+            gpioa.pa2,
+            gpioa.pa3,
             serial::Config::default()
                 .baudrate(115_200.bps()),
             &mut rcc,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -264,15 +264,15 @@ opt-level = "s"
 
 [[example]]
 name = "adc_cont"
-required-features = ["stm32l0x2"]
+required-features = ["stm32l0x2", "io-STM32L071"]
 
 [[example]]
 name = "adc_multi"
-required-features = ["stm32l0x2"]
+required-features = ["stm32l0x2", "io-STM32L071"]
 
 [[example]]
 name = "adc_trig"
-required-features = ["stm32l0x2"]
+required-features = ["stm32l0x2", "io-STM32L071"]
 
 [[example]]
 name = "aes_ecb"
@@ -312,15 +312,19 @@ required-features = ["rt","stm32l0x2"]
 
 [[example]]
 name = "rtc"
-required-features = ["stm32l0x2"]
+required-features = ["stm32l0x2", "io-STM32L071"]
+
+[[example]]
+name = "serial"
+required-features = ["stm32l0x2", "io-STM32L071"]
 
 [[example]]
 name = "serial_dma"
-required-features = ["stm32l0x2"]
+required-features = ["stm32l0x2", "io-STM32L071"]
 
 [[example]]
 name = "serial_dma_async"
-required-features = ["stm32l0x2"]
+required-features = ["stm32l0x2", "io-STM32L071"]
 
 [[example]]
 name = "timer"
--- a/examples/serial.rs
+++ b/examples/serial.rs
@@ -22,20 +22,14 @@ fn main() -> ! {
     // the RCC register.
     let gpioa = dp.GPIOA.split(&mut rcc);
 
-    #[cfg(feature = "stm32l0x1")]
-    let tx_pin = gpioa.pa9;
-    #[cfg(feature = "stm32l0x1")]
-    let rx_pin = gpioa.pa10;
-
-    #[cfg(any(feature = "stm32l0x2", feature = "stm32l0x3"))]
-    let tx_pin = gpioa.pa14;
-    #[cfg(any(feature = "stm32l0x2", feature = "stm32l0x3"))]
-    let rx_pin = gpioa.pa15;
+    // Choose TX / RX pins
+    let tx_pin = gpioa.pa2;
+    let rx_pin = gpioa.pa3;
 
     // Configure the serial peripheral.
     let serial = dp
         .USART2
-        .usart((tx_pin, rx_pin), serial::Config::default(), &mut rcc)
+        .usart(tx_pin, rx_pin, serial::Config::default(), &mut rcc)
         .unwrap();
 
     let (mut tx, mut rx) = serial.split();
--- a/examples/serial_dma.rs
+++ b/examples/serial_dma.rs
@@ -42,7 +42,8 @@ fn main() -> ! {
     let (mut tx, mut rx) = dp
         .USART2
         .usart(
-            (gpioa.pa2, gpioa.pa3),
+            gpioa.pa2,
+            gpioa.pa3,
             serial::Config::default().baudrate(115_200.bps()),
             &mut rcc,
         )
--- a/examples/serial_dma_async.rs
+++ b/examples/serial_dma_async.rs
@@ -46,14 +46,15 @@ fn main() -> ! {
     let gpioa = dp.GPIOA.split(&mut rcc);
 
     let (tx, rx) = dp
-    .USART2
-    .usart(
-        (gpioa.pa2, gpioa.pa3),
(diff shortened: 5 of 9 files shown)


examples/rtc.rs
```diff
@@ -45,7 +45,8 @@ fn main() -> ! {
 
     let serial = dp.USART2
         .usart(
-            (gpioa.pa2, gpioa.pa3),
+            gpioa.pa2,
+            gpioa.pa3,
```
dbrgn: Are independent parameters better, or should we accept a `(TX, RX) where TX: TxPin, RX: RxPin` instead?

rnestler: I like the non tuple version better :slightly_smiling_face:

dbrgn: I noticed that there's a precedent for tuples: https://github.com/stm32-rs/stm32f0xx-hal/blob/33fe6da6dcb7a109e6c8a0251e2f96b770a2dec0/src/serial.rs#L297 We don't necessarily have to follow that though, there are other slight differences in the implementation.

dbrgn:
Rebased.

arkorobotics:
Per the discussion in #67, this PR looks good to me. @hannobraun thoughts?

hannobraun:
I've taken a look and it looks good to me. But I didn't check any of this against the data sheets, and don't have time right now.

> Pin mappings verified against datasheets of stm32l071kb, stm32l072vz and stm32l083vz.

Are we sure that's good enough? I'm not volunteering, but it certainly wouldn't hurt if someone bit the bullet and checked it against all of them :-)

dbrgn:
Rebased again.

> Are we sure that's good enough? I'm not volunteering, but it certainly wouldn't hurt if someone bit the bullet and checked it against all of them :-)

Since the previous values were sometimes plain wrong, it should not get any worse :wink: I did double-check all values against the datasheet, however it probably wouldn't hurt if someone else did that.

My approach was going to [the ST page](https://www.st.com/en/microcontrollers-microprocessors/stm32l0-series.html), opening the product selector for each chip family, picking a device with a large pin count package (e.g. LQFP 100) and taking a look at that datasheet.

The values can be checked using the pin definition table, but it's a bit easier using the alternate functions table.

arkorobotics:
Do we want to wait for these values to be verified or should we merge and update later as needed?

examples/rtc.rs
dbrgn: @arkorobotics @hannobraun what do you think about this?

hannobraun: Sorry for the delay. I don't have an opinion at this time. Just a vague preference for separate arguments, although I don't know why.

dbrgn: Ok, let's keep it like this!

rnestler:
I verified with the STM32L01**0**RB (which isn't supported according to the readme). That one is definitely not compatible since it only has USART2 and LPUART1. Also it doesn't have a variant with 100 pins.
But the pin mappings for USART2 an LPUART1 do match.

dbrgn:
Is there anything blocking this PR?

arkorobotics:
> Is there anything blocking this PR?

cc @hannobraun

hannobraun:
I'm fine with this!

dbrgn:
Crap. This is on hold: https://github.com/stm32-rs/stm32l0xx-hal/pull/76#issuecomment-582302285

Apparently not all STM32L0x1 are created equal when considering AFs.

arkorobotics:
@dbrgn Minor conflicts with the last merge, please advise.

arkorobotics:
Not sure how I closed this PR. Reopening for the time being.

dbrgn:
Thanks! Now that #82 is merged, I'll proceed to generate proper mappings for the different GPIO peripheral versions.

dbrgn:
Alternatively, if #83 were implemented first, that would make things much easier.

dbrgn:
@arkorobotics I think I have sorted this out! As suggested in this comment:

> Thanks! Now that #82 is merged, I'll proceed to generate proper mappings for the different GPIO peripheral versions.

...I made the peripheral availability and the pin mappings dependent on the io-* features, similar to I²C in #87. See #87 for more details.

The pin mappings used in this changeset were taken out of the [cube-parse](https://github.com/dbrgn/cube-parse/tree/700893f4f45ad73d0bd735c0d6fd07e775cddd4b) output for the STM32L0 family. The output can be found [here](https://gist.github.com/dbrgn/8fb6d4640631b705b09db66173536251). To re-generate:

    cargo run -- pin_mappings STM32L0 -d /path/to/stm32cubemx/db/mcu/

The long-term solution should probably be a [universal pin mapping file](https://github.com/stm32-rs/stm32f0xx-hal/pull/29) generated by cube-parse, but with the current APIs this would be a major refactoring. Therefore a manual fix must suffice for now.

Closes #88. Fixes #67.

@arkorobotics this is now ready for review! Maybe @hannobraun would also be willing to take a quick look.

arkorobotics:
Nice work! I love the idea of a universal pin mapping file, but as you pointed out it would require major changes. This is great for now :)

Unless @hannobraun has any objects, I'll go ahead and merge this soon.

hannobraun:
I've only had the time to take a short glance, but no objections from me.

dbrgn:
Great, thanks for the review and the merge!

PR #72, opened 2020-01-09
Further EXTI cleanups/enhancements

jonas-schievink:


hannobraun:
(Travis failure is unrelated to this PR, and already fixed in #66)

jonas-schievink:
Yeah, moving EXTI to a wrapper struct makes sense to me. I think the API was this way because there's no need to initialize anything here (unlike for timers, serial, etc.) so the raw EXTI can be immediately used.

PR #76, opened 2020-02-04
Fix alternative function mapping for I2C on PA9/10

rnestler:


arkorobotics:
This modification doesn't appear to match the Alternate Function pin mapping of the STM32L011. Which exact MCU is this in reference to?

dbrgn:
Oh, that's really bad.

This is for the STM32L071KB.



The STM32L011x3/4 is different indeed:



That means that there are major AF differences *within* the STM32L0X1 series :face_with_head_bandage: 

There are also differences for the USART AFs: The STM32L071 has USART4_TX on PA0/AF6, while the STM32L077 has USART2_RX on PA0/AF0 (and no USART4 on PA0/AF6).

DanSGiesbrecht:
@dbrgn do you have any advice on how to restructure to address differences between MCU's? I'm currently developing for both STM32L071 and STM32L051, and am happy to help.

dbrgn:
I think the best long-term solution would be to extract AF mappings from the database shipped with CubeMX. The relevant discussion is probably at https://github.com/stm32-rs/stm32f0xx-hal/pull/29.

arkorobotics:
Sounds like a good path forward given the inconsistencies. That said, I'm open to other ideas.

PR #80, opened 2020-02-16
Set GPIO reset states per datasheet

craigjb:
Currently, the GPIO reset states all default to` Input<Floating>`, but the STM32L0x3 datasheets specify that the pins are in analog mode at reset, except for a couple special pins.  This change sets all the GPIO states per the datasheet on reset.

Normally, this doesn't cause a problem. But, when compiling in release mode with full optimizations, the `into_floating_input()` method call can be optimized out, and your pin won't be in the right state. Or, in my case:

`pub fn new(_pb0: PB0<Input<Floating>>)`

Here, the unused argument consumes the pin so it won't be used elsewhere, but because it's unused, the call to `into_floating_input()` in the caller doesn't get emitted in the optimized output. :(


craigjb:
Ahhh, I'll have to figure out how to fix those CI failures with the SPI HAL module.

craigjb:
Also, it seems like the SPI pin traits should be implemented for all pin modes, not just the reset ones, no?

arkorobotics:
Good catch. The x1 and x2 datasheets also show 'Analog mode (reset state)' as default after reset.

PR #86, opened 2020-03-24
Add support for continous conversions to ADC

hannobraun:
Adds support for continuous conversions via DMA to the ADC API. This is of somewhat limited use (as the example shows), as it's impossible to process the data as fast as the ADC can produce it. It requires support for hardware triggers, to become fully usable.

However, this PR adds some valuable infrastructure, like circular DMA transfers and the DMA buffer code in `adc`. I think it's worth getting it merged, as it's a solid base to build upon (which I intend to do).

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -156,8 +156,8 @@ impl Adc<Ready> {
     ///
     /// Panics, if `buffer` is larger than 65535.
     #[cfg(feature = "stm32l0x2")]
-    pub fn start<Chan, DmaChan, Buf>(mut self,
-        channel:  Chan,
+    pub fn start<DmaChan, Buf>(mut self,
+        channels: impl Into<Channels>,
         trigger:  Option<Trigger>,
         dma:      &mut dma::Handle,
         dma_chan: DmaChan,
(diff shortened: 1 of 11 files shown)


hannobraun:
There was a compiler warning when building for STM32L0x1/STM32L0x3. Fixing this involved dropping a commit, hence the force-push.

hannobraun:
CI failed due to an unrelated compiler warning. Pushed another commit to fix it.

hannobraun:
I added support for hardware triggers. Since this pull request hasn't been reviewed yet, I figured I'd just push the new commits here instead of creating a new pull request.

This addition makes the new API fully useful.

hannobraun:
And two more commits, adding support for multi-channel conversions. I'm done now, I promise :-)

hannobraun:
CI was boken on non-STM32L0x2 targets. Pushed a commit that should fix that.

src/adc.rs
```diff
@@ -98,6 +135,126 @@ impl Adc {
...
+        dma_chan: DmaChan,
+        buffer:   Pin<Buf>,
+    )
+        -> Adc<Active<DmaChan, Buf>>
+        where
+            DmaToken:    dma::Target<DmaChan>,
+            Buf:         DerefMut + 'static,
+            Buf::Target: AsMutSlice<Element=u16>,
+            DmaChan:     dma::Channel,
+    {
```
dbrgn: Sorry for bikeshedding (I'm not qualified enough to review the actual logic), but do we apply rustfmt to this project?

hannobraun: No, not as a general rule.

arkorobotics:
What a great feature! Thanks so much!

PR #87, opened 2020-03-28
Update I2C pin mappings based on STM32L0xx GPIO IP

### LINKED ISSUE #77: I2C Example - Error Reading Variable
I'm having difficulty implementing the [I2C example](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/examples/i2c.rs):

- MCU: STM32L071
- stm32l0xx_hal version: 0.5.0

```
#![no_std]
#![no_main]

extern crate panic_semihosting;

use cortex_m_rt::entry;
extern crate stm32l0xx_hal;
use stm32l0xx_hal::{pac, prelude::*, rcc::Config};

#[entry]
fn main() -> ! {
    let periph = pac::Peripherals::take().unwrap();

    let mut rcc = periph.RCC.freeze(Config::hsi16());
    let gpioa = periph.GPIOA.split(&mut rcc);

    let sda = gpioa.pa10.into_open_drain_output();
    let scl = gpioa.pa9.into_open_drain_output();

    let mut i2c = periph.I2C1.i2c(sda, scl, 100.khz(), &mut rcc);

    let mut buffer = [0u8; 2];
    const ADDRESS: u8 = 0xFF;

    i2c.write(ADDRESS, &mut buffer).unwrap();

    loop {
    }
}
```

---

When debugging using GDB, I get one out of a variety of errors when I stop debugging, such as:

```
^C
Program received signal SIGINT, Interrupt.
0x080009d8 in stm32l0::stm32l0x1::i2c1::isr::<impl stm32l0::generic::R<u32,stm32l0::generic::Reg<u32,stm32l0::stm32l0x1::i2c1::_ISR>>>::txe (
    self=<error reading variable: Cannot access memory at address 0x20004e3c>)
    at /home/gies/.cargo/registry/src/github.com-1ecc6299db9ec823/stm32l0-0.9.0/src/stm32l0x1/i2c1/isr.rs:801
801	        TXE_R::new((self.bits & 0x01) != 0)
```
```
^C
Program received signal SIGINT, Interrupt.
0x080009e8 in stm32l0xx_hal::i2c::I2c<I,SDA,SCL>::send_byte (
    self=<error reading variable: Cannot access memory at address 0x20004dd4>, 
    byte=<error reading variable: Cannot access memory at address 0x20004dd8>)
    at /home/gies/.cargo/registry/src/github.com-1ecc6299db9ec823/stm32l0xx-hal-0.5.0/src/i2c.rs:183
183	        while self.i2c.isr.read().txe().bit_is_clear() {}
```
```
^C
Program received signal SIGINT, Interrupt.
core::cell::UnsafeCell<T>::get (
    self=<error reading variable: Cannot access memory at address 0x20004db0>)
    at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/cell.rs:1539
1539	/rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/cell.rs: No such file or directory.
```
```
^C
Program received signal SIGINT, Interrupt.
0x080009bc in vcell::VolatileCell<T>::get (
    self=<error reading variable: Cannot access memory at address 0x20004e38>)
    at /home/gies/.cargo/registry/src/github.com-1ecc6299db9ec823/vcell-0.1.2/src/lib.rs:32
32	        unsafe { ptr::read_volatile(self.value.get()) }
```

---

I'm relatively new to Rust. Is this an issue with the I2C example, or am I missing something else? Any assistance would be greatly appreciated.

### LINKED ISSUE #85: I2C 'Wait until byte is transferred' loop
I was reading the `i2c` code since I want to implement it for another chip. Then I encountered the following loop in the `send_byte` function:

```
        // Wait until byte is transferred
        loop {
            let isr = self.i2c.isr.read();
            if isr.berr().bit_is_set() {
                self.i2c.icr.write(|w| w.berrcf().set_bit());
                return Err(Error::BusError);
            } else if isr.arlo().bit_is_set() {
                self.i2c.icr.write(|w| w.arlocf().set_bit());
                return Err(Error::ArbitrationLost);
            } else if isr.nackf().bit_is_set() {
                self.i2c.icr.write(|w| w.nackcf().set_bit());
                return Err(Error::Nack);
            }
            return Ok(());
        }
```
https://github.com/stm32-rs/stm32l0xx-hal/blob/f3b3e824252abfe6e5a592f3ce850d47d56cd112/src/i2c.rs#L188

This loop seems wrong as it only ever runs once. Guess it might still work in practice most of the time. I guess the idea was to wait for `txe` bit to clear like in two lines above? https://github.com/stm32-rs/stm32l0xx-hal/blob/f3b3e824252abfe6e5a592f3ce850d47d56cd112/src/i2c.rs#L183

dbrgn:
The GPIO internal peripheral (IP) does not match the stm32l0x/stm32l0x2/stm32l0x3 grouping used in the PACs. Instead, there are 4 different GPIO IPs in the STM32L0 family that define how the pin functions are mapped to the actual pins. This means that the current I2C mappings are broken for a lot of MCUs. For reference, these issues have already been opened because the mappings were broken:

- https://github.com/stm32-rs/stm32l0xx-hal/pull/76
- https://github.com/stm32-rs/stm32l0xx-hal/issues/77
- https://github.com/stm32-rs/stm32l0xx-hal/issues/85

We can use the `io-*` Cargo features to differentiate between these mappings. This requires that the proper `io-*` feature is set when compiling (and is thus a breaking change). The easiest way to apply the feature without looking at the STM32CubeMX database XML files is to simply use the proper `mcu-*` feature, for example `mcu-STM32L071KBTx`.

The pin mappings used in this changeset were taken out of the [cube-parse](https://github.com/dbrgn/cube-parse/tree/7c369c6bb0c9665decdf575facb751e8463977c7) output for the STM32L0 family. The output can be found [here](https://gist.github.com/dbrgn/c92e2e0641924968d31195611712620b). To re-generate:

    cargo run -- pin_mappings STM32L0 -d /path/to/stm32cubemx/db/mcu/

The source files in the STM32CubeMx database used to generate the mappings:

- /path/to/stm32cubemx/db/mcu/IP/GPIO-STM32L021_gpio_v1_0_Modes.xml
- /path/to/stm32cubemx/db/mcu/IP/GPIO-STM32L031_gpio_v1_0_Modes.xml
- /path/to/stm32cubemx/db/mcu/IP/GPIO-STM32L051_gpio_v1_0_Modes.xml
- /path/to/stm32cubemx/db/mcu/IP/GPIO-STM32L071_gpio_v1_0_Modes.xml

More information on the database format can be found [in the cube-parse README](https://github.com/dbrgn/cube-parse#the-stm32cubemx-database).

The long-term solution should probably be a [universal pin mapping file](https://github.com/stm32-rs/stm32f0xx-hal/pull/29) generated by cube-parse, but with the current APIs this would be a major refactoring. Therefore a manual fix must suffice for now.

Fixes #76, fixes #77, fixes #85.

CC @rnestler @DanSGiesbrecht @ijager

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -28,13 +28,9 @@ use as_slice::AsSlice;
 
 use crate::{
     adc,
-    i2c,
     pac::{
         self,
         dma1::ch::cr,
-        I2C1,
-        I2C2,
-        I2C3,
     },
     rcc::Rcc,
 };
@@ -48,15 +44,11 @@ use crate::pac::USART1;
     feature = "io-STM32L051",
     feature = "io-STM32L071",
 ))]
-use crate::pac::USART2;
-
-#[cfg(any(
-    feature = "io-STM32L021",
-    feature = "io-STM32L031",
-    feature = "io-STM32L051",
-    feature = "io-STM32L071",
-))]
-use crate::serial;
+use crate::{
+    i2c,
+    serial,
+    pac::{I2C1, I2C2, I2C3, USART2},
+};
 
 #[cfg(feature = "stm32l082")]
 use crate::aes;
@@ -563,6 +555,12 @@ impl_target!(
 );
 
 #[cfg(feature = "stm32l0x2")]
+#[cfg(any(
+    feature = "io-STM32L021",
+    feature = "io-STM32L031",
+    feature = "io-STM32L051",
+    feature = "io-STM32L071",
+))]
 impl_target!(
     // I2C1
     i2c::Tx<I2C1>, Channel2, 6;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,8 @@ stm32l082 = ["stm32l0x2"]
 
 # Features based on the GPIO peripheral version.
 # This determines the pin function mapping of the MCU.
+# Note: The easiest way to pick the proper io-* feature is to apply
+# the matching mcu-* feature for your MCU!
 io-STM32L021 = []
 io-STM32L031 = []
 io-STM32L051 = []
@@ -303,8 +305,12 @@ name = "flash"
 required-features = ["rt","stm32l082"]
 
 [[example]]
+name = "i2c"
+required-features = ["rt","stm32l0x2","io-STM32L071"]
+
+[[example]]
 name = "i2c_dma"
-required-features = ["rt","stm32l0x2"]
+required-features = ["rt","stm32l0x2","io-STM32L071"]
 
 [[example]]
 name = "rng"
--- a/build.rs
+++ b/build.rs
@@ -22,7 +22,7 @@ fn main() {
     }
 
     if feature_count != 1 {
-        panic!("\n\nMust select exactly one package for linker script generation!\nChoices: 'stm32l0x1' or 'stm32l0x2' or 'stm32l0x3'\n\n");
+        panic!("\n\nMust select exactly one package for linker script generation!\nChoices: 'stm32l0x1' or 'stm32l0x2' or 'stm32l0x3'\nAlternatively, pick the mcu-feature that matches your MCU, for example 'mcu-STM32L071KBTx'\n\n");
     }
 
     if !cfg!(feature = "disable-linker-script") {
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -22,31 +22,49 @@ use crate::dma::{
     self,
     Buffer
 };
-use crate::gpio::gpioa::{PA10, PA9};
-use crate::gpio::gpiob::{PB6, PB7};
-use crate::gpio::{AltMode, OpenDrain, Output};
 use crate::pac::{
     i2c1::{
         RegisterBlock,
         cr2::RD_WRN_A,
     }
 };
-pub use crate::pac::I2C1;
 use crate::rcc::Rcc;
 use crate::time::Hertz;
 use cast::u8;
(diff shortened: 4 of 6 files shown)


dbrgn:
I have verified this PR against an STM32L071KBTx MCU and at least I2C1 seems to work properly.

arkorobotics:
@dbrgn Looks like #68 has a few merge conflicts with this PR

dbrgn:
> @dbrgn Looks like #68 has a few merge conflicts with this PR

Yup, I just rebased the branch. Let's hope I resolved the conflicts correctly & the tests pass again :slightly_smiling_face:

src/dma.rs
```diff
@@ -563,6 +555,12 @@ impl_target!(
 );
 
 #[cfg(feature = "stm32l0x2")]
+#[cfg(any(
+    feature = "io-STM32L021",
+    feature = "io-STM32L031",
+    feature = "io-STM32L051",
+    feature = "io-STM32L071",
+))]
```
dbrgn, line 563: Multiple `cfg` attributes are additive (so both need to match), right?

I'm not sure whether `stm32l0x2` is still required, but on the other hand I don't have much experience with DMA yet.

arkorobotics:
Looks good! Thanks!

PR #88, opened 2020-04-03
Added all available USART2 pins

mcopela:
The serial file only has one set of USART2 pins under the STM32l0x1 feature flag, while the STM32l031 [setting sheet](https://www.st.com/resource/en/datasheet/stm32l031k6.pdf) lists four. 

I added all the available USART2 pins and tested on my Nucleo-L031K6 evaluation board.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -3,6 +3,7 @@ use core::marker::PhantomData;
 use core::ptr;
 
 use crate::gpio::gpioa::*;
+use crate::gpio::gpiob::*;
 use crate::gpio::{PinMode, AltMode};
 use crate::hal;
 use crate::hal::prelude::*;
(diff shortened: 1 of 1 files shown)


dbrgn:
See #68, which also does this (I think), but is probably wrong as well for some devices in the STM32l0x1 family.

I'm working on a PR that fixes this for all MCUs by basing the cfg flags on the `io-*` features, just like #87 but for UART.

dbrgn:
This should now be fixed in #68.

@mcopela testing that branch would be very welcome! It also includes an API change, where the `Pins` trait is replaced with `RxPin` and `TxPin` (as discussesd in #67).

arkorobotics:
@dbrgn Shall we close this PR given #68 is complete/merged?

dbrgn:
@arkorobotics I would say so, #68 includes the changes that were proposed here!

arkorobotics:
Copy!

mcopela:
Thank you guys! I'll test the new merges today

PR #90, opened 2020-04-05
docs.rs: build for thumbv6m-none-eabi target

jonas-schievink:
Should fix the build on docs.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ version = "0.6.0"
 
 [package.metadata.docs.rs]
 features = ["stm32l0x2", "rt", "stm32-usbd"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -25,6 +25,7 @@ version = "0.6.0"
 
 [package.metadata.docs.rs]
 features = ["stm32l0x2", "rt", "stm32-usbd"]
```
dbrgn, line 27: Slightly offtopic, but you should probably include an `io` feature (e.g. `io-STM32L071`) as well, otherwise - if I understand this correctly - the `serial` and `i2c` modules will be missing from the docs.

Issue #92, opened 2020-04-06
Typo in serial.rs


tytouf:
I encountered what seems to be a typo in serial.rs when building with feature "io-STM32L021"
```
index 190c40d..7b46ea9 100755
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -173,7 +173,7 @@ impl_pins!(
     PA0, AF6, LPUART1, RxPin;
     PA1, AF6, LPUART1, TxPin;
     PA2, AF4, USART2, TxPin;
-    PA2, AF6, LPUART1, TxPin,
+    PA2, AF6, LPUART1, TxPin;
     PA3, AF4, USART2, RxPin;
     PA3, AF6, LPUART1, RxPin;
     PA4, AF6, LPUART1, TxPin;
```

dbrgn:
That does indeed look like a typo! We should make sure that all 4 IO features are tested in CI :confused:

dbrgn:
I'll provide a fix for this case and the CI.

tytouf:
Thank!

tytouf:
On a separate note I was able to toggle a pin and output on the uart very quickly: good job and thank you for the work on this crate!

dbrgn:
This should now be fixed on master, so the issue can be closed.

@arkorobotics maybe there should be a patch release on crates.io since 0.6.0 breaks compilation for some targets?

arkorobotics:
Done! Released v0.6.1.

PR #94, opened 2020-04-06
Travis: Test all io-* features in build matrix

dbrgn:
This should prevent #92 from happening again, by including all four `io-*` features in CI.

If all works well, some tests in this PR should fail.

If that is confirmed, merging #93 and #95 and rebasing should make the tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,10 @@ rust:
   - stable
   - nightly
 env:
-  - MCU=stm32l0x1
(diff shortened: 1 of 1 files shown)


dbrgn:
Tests fail as expected, good! I'll now rebase on #93 to ensure they'll pass afterwards.

dbrgn:
Rebased, this should™ pass now :slightly_smiling_face:

arkorobotics:
Thanks!

PR #96, opened 2020-04-22
pwm: Improve usability of HAL pwm

almusil:
- Add start, stop method that allow user to run
timer on demand

- Allow user to update frequency without the need to
drop the timer instance

- Allow user to free the pac timer in order to properly
use it elsewhere


almusil:
@arkorobotics Can you please review this change?

arkorobotics:
Nice work, thanks for the PR!

PR #98, opened 2020-05-03
i2c: properly handle bus errors

hlukasz:
In send_byte() wait for TX fifo empty before returning Ok. It would be
better to wait for end of transmission, but there is no such field in
I2C status register.

In recv_byte() return Err if error condition detected while waiting for
byte.

Before starting new transaction ensure that both TX and RX buffers are
empty. There are some corner cases where something may be there.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -198,6 +198,12 @@ where
     }
     
     fn start_transfer(&mut self, addr: u8, len: usize, direction: RD_WRN_A, autoend : AUTOEND_A) {
+        // Ensure that TX/RX buffers are empty
+        self.i2c.isr.write(|w| w.txe().set_bit());
+        while self.i2c.isr.read().rxne().bit_is_set() {
+            self.i2c.rxdr.read();
+        };
+        
         self.i2c.cr2.write(|w|
             w
                 // Start transfer


src/i2c.rs
```diff
@@ -217,13 +223,26 @@ where
...
             }
-            return Ok(());
         }
     }
 
     fn recv_byte(&self) -> Result<u8, Error> {
-        while self.i2c.isr.read().rxne().bit_is_clear() {}
+        while self.i2c.isr.read().rxne().bit_is_clear() {
+            let isr = self.i2c.isr.read();
+            if isr.berr().bit_is_set() {
```
almusil, line 235: What about reusing the whole error handling block? send_byte uses almost the same thing. It can be separated to the function.

hannobraun:
Thank your for the PR, @hlukasz, and thanks for the review, @almusil.

I just merged a more recent pull request (#102) that conflicts with this one. Sorry about that! Can you rebase, @hlukasz?

src/i2c.rs
hannobraun: This has been done in #102, I think, so rebase should take care of this comment.

hlukasz:
Done. Basically only one change left after rebasing, but I think that it is still work to add.

hannobraun:
Thank you, @hlukasz!

PR #102, opened 2020-05-10
Made improvements to polling i2c driver

apgoetz:
Hi, I discovered a couple of issues with the blocking I2C driver while playing around with an 32L0538DISCOVERY board. This board has a custom I2C slave peripheral to measure the current of the main STM32 controller, and while writing a driver for it I stumbled upon some bugs in the HAL driver. 

Could someone take a look at these changes and see if they can make it into the library?

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -24,7 +24,7 @@ use crate::dma::{
 use crate::pac::{
     i2c1::{
         RegisterBlock,
-        cr2::RD_WRN_A,
+        cr2::{RD_WRN_A, AUTOEND_A},
     }
 };
 use crate::rcc::Rcc;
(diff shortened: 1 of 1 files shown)


src/i2c.rs
```diff
@@ -193,37 +208,29 @@ where
...
-                self.i2c.icr.write(|w| w.arlocf().set_bit());
-                return Err(Error::ArbitrationLost);
-            } else if isr.nackf().bit_is_set() {
-                self.i2c.icr.write(|w| w.nackcf().set_bit());
-                return Err(Error::Nack);
-            }
-            return Ok(());
-        }
+	// check for any errors
+	return self.check_errors();
```
almusil, line 226: No need for return here

src/i2c.rs
```diff
@@ -181,8 +181,23 @@ where
...
+            self.i2c.icr.write(|w| w.berrcf().set_bit());
+            return Err(Error::BusError);
+        } else if isr.arlo().bit_is_set() {
+            self.i2c.icr.write(|w| w.arlocf().set_bit());
+            return Err(Error::ArbitrationLost);
+        } else if isr.nackf().bit_is_set() {
+            self.i2c.icr.write(|w| w.nackcf().set_bit());
+	    return Err(Error::Nack);
+        }
+        return Ok(());
```
almusil, line 197: No need for return, in fact the whole if can be transformed to not include return statement.

hannobraun:
> Please reformat your change with cargo fmt.

This isn't a policy in this crate, and there's lots of existing code that doesn't follow this convention. I'm open to changing that, but someone would have to put in the legwork. Specifically, I'd like to see two things:
1. Enforce formatting in CI.
2. Adapt the existing code, so `cargo fmt` doesn't completely mess up readability (see https://github.com/nrf-rs/nrf-hal/commit/88b6f8793da4818cdba99f04163195250554fb14, for example).

almusil:
> > Please reformat your change with cargo fmt.
> 
> This isn't a policy in this crate, and there's lots of existing code that doesn't follow this convention. I'm open to changing that, but someone would have to put in the legwork. Specifically, I'd like to see two things:
> 
>     1. Enforce formatting in CI.
> 
>     2. Adapt the existing code, so `cargo fmt` doesn't completely mess up readability (see [nrf-rs/nrf-hal@88b6f87](https://github.com/nrf-rs/nrf-hal/commit/88b6f8793da4818cdba99f04163195250554fb14), for example).

I see. Enforcing it in CI and reformatting the current code is doable. The second part is bit of work but I can create a PR for that if you want. I would like to see the fmt rules enforced.

hannobraun:
> The second part is bit of work but I can create a PR for that if you want.

To be clear, I'm not a fan of rustfmt and would prefer if things stayed as they ar. I can see that I'm in the minority though, and don't intend to stand in the way of progress

So I'm not asking you (or anyone) to do this. I'm just saying, if it's going to be done, that's what it takes to get me to agree :-)

Issue #104, opened 2020-05-18
Faulty DMA Buffer Overflow in adc_trig example


apgoetz:
Hi, 

While experimenting with the ADC DMA examples, I discovered a bug in the implementation. I haven't completely root caused it, but I wanted to file an issue to keep track of it. 

I assume this bug is part of the DMA implementation in #86  

Currently, the DMA peripheral is set up in circular buffer mode to repeatedly transfer ADC samples into a fixed buffer. Sometimes, when the DMA peripheral reaches the end of this buffer, it causes a spurious `BufferOverflow` error when using the `read_available()` method. 

This only seems to occur when the DMA transfer has reached the end of the buffer space and is about to wrap around to the start. The example adc_trig.rs uses a buffer size of 256 and a sample rate of 1Hz, so it takes a while to surface this bug. If you speed up the sample rate or decrease the buffer size, you can reproduce this issue much more quickly. 

The issue seems to be caused by the function `check_overrun_inner()` which is responsible for checking, based on the current DMA status, if the DMA peripheral has overwritten data that has not been read yet. 

It seems that if the DMA write pointer is located at position 0, but the `transfer_state.complete` flag is not set, this causes a spurious buffer overflow. 

The DMA write pointer is calculated from the `DMA_CNDTRx.NDT`  register field, which is supposed to indicate the number of remaining items to transfer. It seems that if this counter has restarted, and is equal to the length of the buffer, but the transfer complete flag is not set, this error condition is triggered. 

I am not certain how this can happen: I assume that the transfer complete flag gets triggered by hardware pretty close to when the `NDT` register reloads its value, but somehow the example code is triggering this behaviour.

apgoetz:
Forgot to mention, I am using a 32L0538DISCOVERY dev board, which is an stm32l0x3 device. In order to get the example to build, I removed the `#cfg` checks in `adc.rs`, and the required features from `Cargo.toml`. 

I also had to change the UART peripheral to match the dev board, but I assume these changes shouldn't affect the behavior of the code. 

The ADC DMA transfers seem to be working, until the example code eventually panics due to the aforementioned BufferOverflow error.

dbrgn:
(Offtopic, but what did you need to adjust with the UART peripheral? Did you specify the `mcu-STM32L053C8Tx` cargo feature?)

hannobraun:
Thanks for the report! I'll take a look later to see if I can reproduce this with the STM32L0x2 device I developed this with.

[...]

PR #105, opened 2020-05-27
New ADC DMA implementation

apgoetz:
I'd like to get some feedback on a new DMA driver I have been working on. 

This exposes an interface that allows for one shot dma sessions, where a whole block of samples are transferred at once, but the DMA session ends after this block has completed.

The current DMA implementation sets up a continuous DMA transfer, which makes it hard to reason about the soundness of the buffer, since it is possible the DMA controller could be writing to the memory that is accessed by the CPU. 

I have left the existing DMA implementation untouched in this branch. I would like some feedback on this approach, to see if the APIs that are exposed seem reasonable. 

I have also only updated one example to use this new code, it is in adc_dma.rs. The other dma examples all use the existing interface. These will need to be updated before this branch can merge.


hannobraun:
Thanks, @apgoetz, I'll take a look later. Feel free to ping me, if it takes too long.

src/adc.rs
```diff
@@ -535,6 +537,11 @@ impl Channels {
     {
         self.flags |= 0x1 << C::channel()
     }
+
+    /// constructor for empty channels object
+    pub fn new() -> Self {
```
hannobraun, line 542: I think it would be nicer to have have the constructor come first in the `impl` block, but that's not important.

src/adc.rs
```diff
@@ -547,6 +554,56 @@ impl<C> From<C> for Channels
...
+    fn into_channels(&self) -> Channels {
+	self.clone()
+    }
+}
+
+impl<C> IntoChannels for C
+    where C: Channel<Adc<Ready>, ID=u8>
+{
+    fn into_channels(&self) -> Channels {
+        Channels{flags: 0x1 << C::channel()}
```
hannobraun, line 575: Might be nicer to re-use `add` here. The compiler should be able to optimize that (if it doesn't, it might be possible to convince it by adding `#[inline(always)]` to `add`).

examples/adc_trig.rs
```diff
@@ -27,6 +27,10 @@ use stm32l0xx_hal::{
     serial,
 };
 
+use stm32l0xx_hal::serial::Serial1Ext;
+
+const BUFSIZE : usize = 8; 	// the size of the buffer to use
+const FREQUENCY : u32 = 1; // the frequency to sample at
```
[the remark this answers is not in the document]
hannobraun: Regarding all the changes that seem to be in the wrong commit: Fixing this is probably going to be a pain and will matter very little after this is merged. Might not worth the effort.

hannobraun:
Closing for inactivity and because #127 provided an alternative fix for #104.

@apgoetz, feel free to re-open (or open a new pull request), if you want to pick this back up.

PR #107, opened 2020-05-27
Enforce rust fmt

almusil:
@hannobraun Can you please take a look?

### DIFF
--- a/src/pwr.rs
+++ b/src/pwr.rs
@@ -2,22 +2,12 @@
 //!
 //! See STM32L0x2 reference manual, chapter 6.
 
-
-use cortex_m::{
-    asm,
-    peripheral::SCB,
-};
+use cortex_m::{asm, peripheral::SCB};
 
 use crate::{
     pac,
-    rcc::{
-        Clocks,
-        ClockSrc,
-        PLLSource,
-        Rcc,
-    },
+    rcc::{ClockSrc, Clocks, PLLSource, Rcc},
 };
-
 
 /// Entry point to the PWR API
 pub struct PWR(pac::PWR);
@@ -54,7 +44,9 @@ impl PWR {
         while self.0.csr.read().vosf().bit_is_set() {}
 
         // Safe, as `VcoreRange` only provides valid bit patterns.
-        self.0.cr.modify(|_, w| unsafe { w.vos().bits(range as u8) });
+        self.0
+            .cr
+            .modify(|_, w| unsafe { w.vos().bits(range as u8) });
 
         while self.0.csr.read().vosf().bit_is_set() {}
     }
@@ -111,10 +103,7 @@ impl PWR {
 
     /// Returns a struct that can be used to enter Sleep mode
     pub fn sleep_mode<'r>(&'r mut self, scb: &'r mut SCB) -> SleepMode<'r> {
-        SleepMode {
-            pwr: self,
-            scb,
-        }
+        SleepMode { pwr: self, scb }
     }
 
     /// Returns a struct that can be used to enter low-power sleep mode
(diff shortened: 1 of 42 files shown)


dbrgn:
I'm definitely in favor of using and enforcing rustfmt everywhere.

Not sure whether this should be a single reformatting commit though. Is there any advantage in having multiple commits?

almusil:
IMO it is better to keep the history what was actually done. But I don't mind squashing it into single commit.

dbrgn:
I'm a big fan of keeping the history clean, but in "git log FILE" and "git blame FILE" both approaches will show up as a single commit :slightly_smiling_face: In contrast, I tend to find a single commit easier to "parse" in the commit history than 17. (The CI change makes sense as a separate commit though.)

In any case, no strong opinons on that. I do think enforcing rustfmt is a good idea though, it completely eliminates the opinionated "could you please format this part of the code in a different way" from PR reviews.

src/pwr.rs
```diff
@@ -308,61 +284,72 @@ impl PowerMode for StopMode<'_> {
...
-            match self.rcc.clocks.source() {
+        self.rcc
+            .rb
+            .cfgr
+            .modify(|_, w| match self.rcc.clocks.source() {
                 ClockSrc::MSI(_) =>
-                    // Use MSI as clock source after wake-up
-                    w.stopwuck().clear_bit(),
+                // Use MSI as clock source after wake-up
+                {
```
dbrgn, line 293: This one looks a bit weird... Since when does rustfmt put opening curly braces on its their own line? Might be a mess-up due to the comment...

almusil: It probably is. I will rearrange the comment so it does not look that bad. Good catch thanks

src/pwr.rs
```diff
@@ -308,61 +284,72 @@ impl PowerMode for StopMode<'_> {
...
+                // panic, which is also documented in this method's doc
+                // comment.
+                {
+                    panic!("External clock not supported for Stop mode")
+                }
+            });
 
         // Configure Stop mode
-        self.pwr.0.cr.modify(|_, w|
+        self.pwr.0.cr.modify(|_, w| {
```
almusil, line 338: I see that I forgot to change this one. On it

hannobraun:
Thank you, @almusil and @dbrgn! The formatting changes look good to me (well, as good as it gets with rustfmt'ed code, at least :-) ).

I see two minor problems with the last commit:
1. I think formatting should not be enforced on nightly. I think that has lead to problems in the past in other projects, but I don't remember clearly why that was. Probably, nightly rustfmt tends to make different formatting choices than stable/beta, which means it's going to fail all the time anyway.
2. It would be great if there was a script that was called by CI, that I could also run locally to figure out whether CI was likely to pass ([example of this in another HAL](https://github.com/lpc-rs/lpc8xx-hal/blob/master/scripts/build.sh)). Otherwise, people are going to submit unformated code and won't even know right away, because CI takes a while to run and GitHub won't notify the PR author if it fails. This is going to lead to a lot of unnecessary back-and-forth.

I think it's fine to merge as-is though, and follow up with additional improvements later, as necessary. Especially point 2 is a bit of a can of worms of worms, due to the use of the build Matrix on CI (that other HAL I linked doesn't use the build Matrix for that reason).

@arkorobotics What do you think? I'm inclined to merge this, to get the formatting discussion over with.

hannobraun:
I've just introduced a conflict by merging #101.

@almusil, can you take a look?

almusil:
@hannobraun  Rebased, thanks

almusil:
> Thank you, @almusil and @dbrgn! The formatting changes look good to me (well, as good as it gets with rustfmt'ed code, at least :-) ).
> 
> I see two minor problems with the last commit:
> 
>     1. I think formatting should not be enforced on nightly. I think that has lead to problems in the past in other projects, but I don't remember clearly why that was. Probably, nightly rustfmt tends to make different formatting choices than stable/beta, which means it's going to fail all the time anyway.
> 

Rust fmt might enforce different rules on nightly and also fmt is not always available there. So it makes sense to skip fmt on nightly. 

>     2. It would be great if there was a script that was called by CI, that I could also run locally to figure out whether CI was likely to pass ([example of this in another HAL](https://github.com/lpc-rs/lpc8xx-hal/blob/master/scripts/build.sh)). Otherwise, people are going to submit unformated code and won't even know right away, because CI takes a while to run and GitHub won't notify the PR author if it fails. This is going to lead to a lot of unnecessary back-and-forth.

This should not be that hard I will create a issue for that and assign myself. Once I have some more spare time I'll look into this.
> 
> 
> I think it's fine to merge as-is though, and follow up with additional improvements later, as necessary. Especially point 2 is a bit of a can of worms of worms, due to the use of the build Matrix on CI (that other HAL I linked doesn't use the build Matrix for that reason).
> 
> @arkorobotics What do you think? I'm inclined to merge this, to get the formatting discussion over with.

almusil:
Any update?

dbrgn:
There's a new conflict, maybe you can resolve that?

almusil:
> There's a new conflict, maybe you can resolve that?
> 

Done

> > Rust fmt might enforce different rules on nightly and also fmt is not always available there. So it makes sense to skip fmt on nightly.
> 
> I think this is still missing as well, right? Or would that be part of #109?

It was meant to be part of #109 but I can probably do it in this patch. Only thing that might make it harder are the conflicts so if it's not vital I would rather do it in follow up.

hannobraun:
Nobody has complained, and this has been open long enough. I'm merging.

Thank you, @almusil!

Issue #114, opened 2020-06-28
Linked Upcounting Timer


dbrgn:
I wrote some code to link two 16 bit timers on an STM32L0 in order to form a 32 bit upcounting timer that can be used to measure relative time. The goal is to use this abstraction in order to implement the `Monotonic` trait for the RTIC scheduler.

Here's the current code, needs more testing & should be made more general (there are other timer pairs that allow linking), but seems to work already:

```rust
/// Two linked 16 bit timers that form a 32 bit upcounter.
pub struct LinkedTimer {
    /// Timer in master mode
    tim2: TIM2,
    /// Timer in slave mode
    tim3: TIM3,
}

impl LinkedTimer {
    /// Create a new `LinkedTimer` with TIM2 as master and TIM3 as slave.
    pub fn new(tim2: TIM2, tim3: TIM3, rcc: &mut Rcc) -> Self {
        // Enable timers
        rcc.rb.apb1enr.modify(|_, w| w.tim2en().set_bit());
        rcc.rb.apb1enr.modify(|_, w| w.tim3en().set_bit());

        // Reset timers
        rcc.rb.apb1rstr.modify(|_, w| w.tim2rst().set_bit());
        rcc.rb.apb1rstr.modify(|_, w| w.tim2rst().clear_bit());
        rcc.rb.apb1rstr.modify(|_, w| w.tim3rst().set_bit());
        rcc.rb.apb1rstr.modify(|_, w| w.tim3rst().clear_bit());

        // Enable counter
        tim2.cr1.modify(|_, w| w.cen().set_bit());
        tim3.cr1.modify(|_, w| w.cen().set_bit());

        // In the MMS (Master Mode Selection) register, set the master mode so
        // that a rising edge is output on the trigger output TRGO every time
        // an update event is generated.
        tim2.cr2.modify(|_, w| w.mms().variant(tim2::cr2::MMS_A::UPDATE));

        // In the SMCR (Slave Mode Control Register), select the internal
        // trigger 0 (ITR0) as trigger source (TS).
        //
        // See table 76 ("TIM2/TIM3 internal trigger connection") in the
        // reference manual RM0377.
        tim3.smcr.modify(|_, w| w.ts().variant(tim2::smcr::TS_A::ITR0));
        // Set the SMS (Slave Mode Selection) register to "external clock mode 1",
        // where the rising edges of the selected trigger (TRGI) clock the counter.
[...]

PR #115, opened 2020-06-28
LinkedTimer: Implement initial version

dbrgn:
Implementation of #114.

A `LinkedTimer` impl is initialized with two hardware timers (either TIM2/TIM3 or TIM21/TIM22). The two timers are configured in master/slave mode so that an overflow of the master timer triggers an update on the slave timer. This way, two 16 bit timers can be combined to a single 32 bit timer. (The STM32L0 does not have 32 bit timers.)

2020-06-29-003656_743x314_scrot

Such a linked timer allows implementing the `Monotonic` trait for the RTIC scheduler with high timer precision.

Right now the implementation does not add logic for detecting an overflow, that is left to the user. Furthermore, no prescaler is configured. And not all possible timer combinations have been implemented.

An example is provided (tested on the STM32L071KBTx). Here's the serial output:

```
Starting example
Init TIM2/TIM3...
Init TIM21/TIM22...
TIM2/TIM3    count   26790878 (msb=408 lsb=52211 Δ26790878 53581 kHz)
TIM21/TIM22  count   11182841 (msb=170 lsb=41743 Δ11182841 22365 kHz)
TIM2/TIM3    count   34790917 (msb=530 lsb=56858 Δ8000039 16000 kHz)
TIM21/TIM22  count   19166170 (msb=292 lsb=29680 Δ7983329 15966 kHz)
TIM2/TIM3    count   42790960 (msb=652 lsb=61509 Δ8000043 16000 kHz)
TIM21/TIM22  count   27166218 (msb=414 lsb=34336 Δ8000048 16000 kHz)
TIM2/TIM3    count   50791003 (msb=775 lsb=624 Δ8000043 16000 kHz)
TIM21/TIM22  count   35132916 (msb=536 lsb=5642 Δ7966698 15933 kHz)
TIM2/TIM3    count   58791046 (msb=897 lsb=5275 Δ8000043 16000 kHz)
...
```

Let me know whether this is useful, and whether the implementation looks correct. This was my first STM32 timer deep-dive, I spent quite a few hours looking through the reference manual to figure out how to link timers :slightly_smiling_face: 

The relevant information in the reference manual (RM0377) is in section 16.3.15 "Timer synchronization". Here's the section that describes timer linking:

2020-06-29-003857_765x358_scrot

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -144,7 +144,8 @@ macro_rules! timers {
                 }
 
                 /// Select master mode
-                pub fn select_master_mode(&mut self,
+                pub fn select_master_mode(
+                    &mut self,
                     variant: <$TIM as GeneralPurposeTimer>::MasterMode,
                 ) {
                     self.tim.select_master_mode(variant);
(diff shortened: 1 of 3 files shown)


dbrgn:
> CI is failing though, due to the RTFM -> RTIC rename. It probably makes sense to fix this in a separate PR, then rebase this one.

Yep, I meant to do that (but it was already late yesterday). I'll provide a PR tonight, afterwards we can rebase.

src/timer.rs
```diff
@@ -225,3 +225,104 @@ pub trait GeneralPurposeTimer {
...
+                /// Return the current 16 bit counter value of the master timer (LSB).
+                fn get_counter_lsb(&self) -> u16 {
+                    self.tim_master.cnt.read().cnt().bits()
+                }
+
+                /// Return the current 32 bit counter value.
+                fn get_counter(&self) -> u32 {
+                    let lsb = self.tim_master.cnt.read().cnt().bits() as u32;
+                    let msb = self.tim_slave.cnt.read().cnt().bits() as u32;
+                    (msb << 16) | lsb
```
timokroeger, line 313: Assuming the timer runs at a high frequency:
1. Read master: master_cnt = 0xFFFF, slave_cnt = 0x003 (but we have not read the slave yet)
2. Read slave: master_cnt = 0x0000,  slave_cnt = 0x004
3. Return 0x0004FFFF even though the actual value should be 0x0003FFF
Not sure what’s the best way to prevent this. Maybe read the slave before and after the master counter and check if it changed

dbrgn: Hmm, good point... I assume there's no way to read two registers simultaneously.

Your approach with a change check seems sensible. But what do we do if the slave changed? Retry the whole process? Return an error? Could it be a problem if reading the timer is not constant time?

dbrgn: I managed to provoke the error quite easily by setting the ARR register in the LSB register to a low value like 0xff.

```
MSB mismatch: 26773 != 26774
```

I pushed a new commit that should fix the race condition.

dbrgn:
Updated, now with reset :slightly_smiling_face: 

I discovered that this cannot be used for RTIC (since the RTIC monotonic timer must be initialized statically), but it may be useful nevertheless.

hannobraun:
Looks like this was left un-merged for no good reason. Merging.

PR #116, opened 2020-06-29
Rename RTFM to RTIC

dbrgn:


Cargo.toml
```diff
@@ -44,7 +44,7 @@ aligned = "0.3.1"
 heapless = "0.5.0"
 panic-halt = "0.2.0"
 panic-semihosting = "0.5.1"
-cortex-m-rtfm = "0.5.1"
+cortex-m-rtic = "0.5.1"
```
hannobraun, line 47: That version doesn't exist. This should work:
```suggestion
cortex-m-rtic = "0.5.3"
```

dbrgn: Oh, good point!

Issue #118, opened 2020-08-01
.i2c method not found in `stm32l0xx_hal::pac::I2C1`


nebelgrau77:
I must be missing something, so please help: I'm trying to convert a working code, using SSD1306 display, from another STM32 board. I got stuck at setting up the I2C. It just won't compile, with the error as in the title. I tried the example from the crate, same problem. And when looking in the docs, I can't, in fact, find the .i2c method. After dp.I2C1. rust-analyzer shows only registers' names as possible methods. Am I missing something in the Cargo.toml that has to be set up?

The board is STM32L031, and the code: 

```rust

    let dp = pac::Peripherals::take().unwrap();
  
    let mut rcc = dp.RCC.freeze(Config::hsi16());

    let gpioa = dp.GPIOA.split(&mut rcc);

    let sda = gpioa.pa10.into_open_drain_output();
    let scl = gpioa.pa9.into_open_drain_output();

    let mut i2c = dp.I2C1.i2c(sda, scl, 100.khz(), &mut rcc);
```

dbrgn:
You probably need to pick the appropriate `io-*` Cargo feature, so that the HAL knows the proper function mapping of the I/O pins. The easiest way to do this is to use the `mcu-*` feature matching your MCU:

https://github.com/stm32-rs/stm32l0xx-hal/blob/d7f0e504bdb42af588cd1456c8e3b901766a0d45/Cargo.toml#L97

For example:

```toml
# Cargo.toml
stm32l0xx_hal = { version = "*", features = ["mcu-STM32L071KBTx"] }
```

(We really need to document this properly. I'll create a PR soon-ish.)

dbrgn:
@nebelgrau77 can you check out the updated README in https://github.com/stm32-rs/stm32l0xx-hal/tree/improve-docs and let us know whether it clears things up?

nebelgrau77:
OK, so that's what I thought, but I can't really get it to work. I did add the io-... to the features, but to no avail. Do I need to copy the whole set from the Cargo.toml in the examples, or is it enough to do this: 
```rust
stm32l0xx-hal = {version = "0.6.2", features = ["io-STM32L031", "stm32l0x1", "lqfp32","rt"]}
```
? 
Unfortunately it didn't work, just tried it :/

dbrgn:
What MCU are you using? The only two features you need are `rt` and `mcu-*` (the latter of course with your concrete MCU).

[...]

Issue #103, opened 2020-05-13
serial doesn't seem to work at all


nebelgrau77:
I'm trying to use the serial example on STM32L031K6 Nucleo board. I'm getting the following compilation errors: 

```cargo
error[E0432]: unresolved import `stm32l0xx_hal::serial`: no `serial` in the root

error[E0599]: no method named `usart` found for struct `stm32l0::stm32l0x1::USART2` in the current scope

.usart(tx_pin, rx_pin, serial::Config::default(), &mut rcc)
          ^^^^^ method not found in `stm32l0::stm32l0x1::USART2`
```

in fact I can't find these in the documentation, but they clearly do exist in the actual src code in the repo. How can I get it to work?

dbrgn:
You probably didn't specify your MCU (see https://github.com/stm32-rs/stm32l0xx-hal/blob/b8147f13c455942b4374c353b33b486e23b30172/Cargo.toml#L97-L258).

For the STM32L031K6 you need `features = ["mcu-STM32L031K6Tx"]` or `features = ["mcu-STM32L031K6Tx"]`, depending on which MCU you have. (If you aren't sure which exact MCU you have, you can also use `features = ["io-STM32L031", "stm32l0x1"]`.)

dbrgn:
We should probably add a note about that in the README...

nebelgrau77:
Thanks! I was just missing the "io-..." part in the features block. Now it's working as expected :)

PR #119, opened 2020-08-01
Improve README

dbrgn:
A lot of people had troubles with using this HAL because the Cargo features were not set up properly. For example:

- https://github.com/stm32-rs/stm32l0xx-hal/issues/118
- https://github.com/stm32-rs/stm32l0xx-hal/issues/103

These features were inadequately documented. The first commit of this PR attempts to fix that by documenting how to properly specify a dependency on `stm32l0xx-hal` (and the available Cargo features).

Additionally, I simplified the toolchain setup docs. The repo still contains helper scripts for flashing and debugging. These scripts are highly dependent on your debug probe (JLink? STLink? Something else?), your toolchain (OpenOCD? JLinkExe? The STLink tool? Cargo Flash?), your OS and your MCU. I do not think they should belong in this repo. Documentation on different setups should probably be moved to the Embedded Rust book. My suggestion would be to remove the helper scripts and the corresponding sections in the README (the README updated by this PR already contains a link to the Embedded Rust book).

If there's a toolchain we should recommend, then in my opinion it should be cargo-flash or cargo-embed (as soon as GDB debugging works reliably). Shoutout to @yatekii @Tiwalun @adamgreig and many others that are working on the probe-rs project.

What do you think, @arkorobotics?

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,51 +78,46 @@ for example the `STM32L062K8Tx` uses the GPIO peripheral version named
 `io-STM32L051`.
 
 
-# Build Dependencies
+# Toolchain Setup
 
-1. Rustup toolchain installer
+In order to use this HAL, you need the following Setup:
 
-    https://rustup.rs
+1. Install Rustup
 
+    See [rustup.rs](https://rustup.rs/) for details. You may als be able to
+    install Rustup directly through your distro.
 
-# Toolchain Configuration
+2. Install the `arm-none-eabi` compiler toolchain
 
-`$ rustup target add thumbv6m-none-eabi`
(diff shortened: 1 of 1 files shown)


arkorobotics:
The Readme is long overdue for an update. The improvements you've made in this PR look good. I just want to make sure we capture the issues with flashing the L0 family MCU's with OpenOCD v.0.10.0. I'll look into adding the instructions (removed in this PR) to the Embedded Rust book as the flashing issue can be frustratingly difficult to track down (and it doesn't seem OpenOCD is planning to release the fixed version any time soon).

That said, I'm going to merge this PR. Thanks!

dbrgn:
Great, thanks for the quick review @arkorobotics :slightly_smiling_face:

PR #121, opened 2020-08-05
Expose factory calibration data

### LINKED ISSUE #120: Register definitions for factory option bytes
I'm trying to use the internal ADC reference to monitor VDDA. To do that I need the `VREFINT_CAL` value from the factory options. The factory option location aren't available in the SVD file (probably because they aren't peripheral) so they are not accessible.

Would it make sense to add register like definitions to the HAL or the PAC for the factory option location?

For reference a section from the STM32L071xx datasheet:

vrefint_cal_location

and from the STM32L0x1 reference manual:

vrefint_cal

cc @dbrgn

dbrgn:
This PR exposes the factory calibration data.

Verified to be working on an STM32L071KBTx. As a cross-check, I verified the memory addresses against the datasheet of the (randomly picked) STM32L053C6, they match.

2020-08-05-233719_346x79_scrot

Strongly inspired by the `signature` module in https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/signature.rs.

Fixes #120.

CC @rnestler

### DIFF
--- a/src/calibration.rs
+++ b/src/calibration.rs
@@ -1,0 +1,88 @@
+//! Factory-programmed calibration data
+//!
+//! The STM32L0 contains a few read-only registers with factory calibration
+//! data that are written during production.
+
+macro_rules! define_ptr_type {
+    ($name: ident, $ptr: expr) => {
+        impl $name {
+            fn ptr() -> *const Self {
+                $ptr as *const _
+            }
+
+            /// Returns a wrapped reference to the value in flash memory
+            pub fn get() -> &'static Self {
+                unsafe { &*Self::ptr() }
+            }
+        }
+    };
+}
+
+/// Internal voltage reference calibration data (VREFINT_CAL)
+///
+/// The internal voltage reference provides a stable (bandgap) voltage output
+/// for the ADC and Comparators. The precise voltage of V_REFINT is
+/// individually measured for each part by ST during production test and stored
+/// in the system memory area. It can be accessed through this struct.
+#[derive(Debug)]
+#[repr(C)]
+pub struct VrefintCal(u16);
+define_ptr_type!(VrefintCal, 0x1FF8_0078);
+impl VrefintCal {
+    /// Read calibration value (VREFINT_CAL), acquired at temperature of 25 °C and V_DDA = 3 V.
+    pub fn read(&self) -> u16 {
+        self.0
+    }
+}
+
+/// Temperature sensor calibration data acquired at 30 °C (TS_CAL1)
+///
+/// The temperature sensor (T_SENSE) generates a voltage (V_SENSE) that varies
+/// linearly with temperature.
+///
+/// The sensor provides good linearity, but it has to be calibrated to obtain
+/// good overall accuracy of the temperature measurement. As the offset of the
+/// temperature sensor varies from chip to chip due to process variation, the
+/// uncalibrated internal temperature sensor is suitable for applications that
+/// detect temperature changes only.
+///
+/// To improve the accuracy of the temperature sensor measurement, each device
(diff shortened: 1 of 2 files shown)


src/calibration.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+    };
+}
+
+/// Internal voltage reference (V_REFINT)
+///
+/// The internal voltage reference provides a stable (bandgap) voltage output
+/// for the ADC and Comparators. The precise voltage of V_REFINT is
+/// individually measured for each part by ST during production test and stored
+/// in the system memory area. It is accessible in read-only mode.
```
hannobraun, line 26: What's "read-only mode"? Does that mean the register can only be read, not written, or is there some mode switching involved that I'm not aware of?

src/calibration.rs
```diff
@@ -0,0 +1,88 @@
...
+/// The sensor provides good linearity but it has to be calibrated to obtain
+/// good overall accuracy of the temperature measurement. As the offset of the
+/// temperature sensor varies from chip to chip due to process variation, the
+/// uncalibrated internal temperature sensor is suitable for applications that
+/// detect temperature changes only.
+///
+/// To improve the accuracy of the temperature sensor measurement, each device
+/// is individually factory-calibrated by ST. The temperature sensor factory
+/// calibration data are stored by ST in the system memory area, accessible in
+/// read-only mode.
```
hannobraun, line 52: See comment about read-only mode above.

src/calibration.rs
```diff
@@ -0,0 +1,88 @@
...
```
hannobraun, line 78: See comment about read-only mode above.

src/calibration.rs
```diff
@@ -0,0 +1,88 @@
...
+/// The internal voltage reference provides a stable (bandgap) voltage output
+/// for the ADC and Comparators. The precise voltage of V_REFINT is
+/// individually measured for each part by ST during production test and stored
+/// in the system memory area. It is accessible in read-only mode.
+#[derive(Debug)]
+#[repr(C)]
+pub struct VrefintCal(u16);
+define_ptr_type!(VrefintCal, 0x1FF8_0078);
+impl VrefintCal {
+    /// Read calibration value, acquired at temperature of 25 °C and V_DDA = 3 V.
```
rnestler, line 32: Maybe mention `VREFINT_CAL` somewhere which is the name used in the datasheet. Makes it easier to search for it in the documentation :slightly_smiling_face:

src/calibration.rs
rnestler: The "read-only mode" is actually copy paste from the datasheet, but I agree that the wording is confusing (as are ST datasheets in general :wink:)

src/calibration.rs
hannobraun: Good point. I actually prefer to follow the naming convention from the official documentation where practical, even if that conflicts with Rust's naming convention. So I wouldn't mind if this struct were called `VREFINT_CAL`, if that's the "official" name.

src/calibration.rs
dbrgn: I don't find it confusing (it is written by ST in the factory, but accessible to the developer in read-only mode) but I can also just remove that part of the sentence, since it doesn't really add anything useful.

src/calibration.rs
dbrgn: I'd prefer to stick to Rust naming for the struct, but I'll add a reference to `VREFINT_CAL` to the docs!

dbrgn:
Docs improved and branch rebased against master!

hannobraun (approved):
Still haven't checked this against the reference manual, but everything looks good to me now. Thank you, @dbrgn!

Issue #123, opened 2020-08-30
ADC-DMA confiuguration improperly feature blocked


electroCutie:
stm32l0x1 chips support continuous sampling and moving values with DMA

This was confirmed on a NUCLEO-L011K board

hannobraun:
Fixed in #124, I believe.

PR #124, opened 2020-08-30
Removed config tests that prevented stm32l0x1 dfevices from using ADC with DMA

electroCutie:
This is a supported configuration and I've tested it with a NUCLEO-L011K development
board and confirmed in the spec sheet that this is supported

Proposed improvement for  #123


hannobraun:
Thank you, @electroCutie! The CI build fails due to formatting (should be fixed by running `cargo fmt`), but otherwise this looks good to me.

Issue #125, opened 2020-08-30
Improper ADC-DMA Overruns


electroCutie:
When calculating position in the buffer using values from the DMA registers the Buffer doesn't account for witnessing 0

I also managed to sometimes and very spuriously see values much larger than the transfer size. They were seeming arbitrary and had no significant bit pattern

electroCutie:
this is a duplicate of #104

PR #126, opened 2020-08-30
fix improper reporting of DMA overruns when remaining bytes is outside the expected range

electroCutie:
I've witnessed zeros as well as values larger than the initial setting,
albeit spuriously. Added a saturating subtraction and a modulo to the
position calculation and overrun now seems to be working correctly

There still seems to be more overruns than should strictly be happening,
but I've been unable to pin them down and they disappear when clocking
low enough. They might even be real but it is difficult to know. But
this fix is solid

Fix for #125

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -482,7 +482,7 @@ impl Buffer {
 
         // Let's translate what we got from the DMA peripheral into a write
         // position that we can compare with our read position.
-        let pos = self.len - remaining;
+        let pos = self.len.saturating_sub(remaining) % self.len;
 
         TransferState {
             pos,
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -482,7 +482,9 @@ impl Buffer {
 
         // Let's translate what we got from the DMA peripheral into a write
         // position that we can compare with our read position.
-        let pos = self.len.saturating_sub(remaining) % self.len;
+        // Zero can be witnessed in remaining before the DMA reloads
+        // the register so we wrap the result of the subtraction around
+        let pos = (self.len - remaining) % self.len;
 
         TransferState {
             pos,


src/adc.rs
```diff
@@ -482,7 +482,7 @@ impl Buffer {
 
         // Let's translate what we got from the DMA peripheral into a write
         // position that we can compare with our read position.
-        let pos = self.len - remaining;
+        let pos = self.len.saturating_sub(remaining) % self.len;
```
hannobraun, line 485: I don't understand what's happening here. First off, it's very weird that the hardware reports values for `remaining` that are too large, but okay, hardware is often quite weird :-)

Still, it's not clear to me what should happen in that case. The saturating sub will prevent an underflow, which is good, but are we certain that it will lead to correct behavior? Maybe there's an errata note somewhere that can shed some light on the hardware behavior.

I also don't understand what the modulo is doing. Will `self.len.saturating_sub(remaining)` not always result in a value that's smaller or equal to `self.len`? In that case the `% self.len` won't do anything, or am I missing something obvious?

hannobraun:
Thank you, @electroCutie. I've left a comment in your commit. Some additional notes:

- There are known problems with ADC/DMA that might account for the overruns you're seeing. See #104 and #105.
- I think it's better to move some of the explanation from your commit message into the code, as a comment. When wondering why something looks like it does, it's always nicer to have a comment right there, instead of having to hunt for the right commit message.

src/adc.rs
```diff
@@ -482,7 +482,9 @@ impl Buffer {
 
         // Let's translate what we got from the DMA peripheral into a write
         // position that we can compare with our read position.
-        let pos = self.len.saturating_sub(remaining) % self.len;
+        // Zero can be witnessed in remaining before the DMA reloads
+        // the register so we wrap the result of the subtraction around
+        let pos = (self.len - remaining) % self.len;
```
hannobraun, line 487: Okay, so the thing I missed before is that if `remaining` is zero, the `% self.len` will result in `pos = 0` instead of `pos = self.len`. Makes sense.

What I don't understand yet is the wider context of this change. Under which circumstances will `remaining == 0` actually cause an overrun, and why? How will this affect overrun detection?

I'm not saying this is wrong, but I'm also not convinced of the following things:
- That this is correct, given the complexity of the surrounding code.
- That the overruns you're seeing are actually caused by `remaining == 0`, given the known issue with ADC/DMA.
- If you're experiencing overruns not caused by #104, that this is the right place to fix them. Maybe it should be fixed in the overrun detection code, for example.

This code is quite complex, and we're really suffering from the lack of a test suite here. I think the best way to get this merged is to do a more thorough analysis and write up the results in a comment in the appropriate place. As it is, I don't understand this fix (and its underlying problem) well enough to merge it.

electroCutie: Yeah. I'll do that. I thought I had a good handle on it but I will review #104 and see if I can really pin this down

I think I misunderstood the bug but now looking at #104 I think this is a duplicate of that bug. Lets close this merge and my version of the bug and i'll focus my efforts there

hannobraun: Sounds good!

Please note that my latest conclusion was, that this API can't be supported given the hardware behavior (https://github.com/stm32-rs/stm32l0xx-hal/issues/104#issuecomment-631383181), and I currently favor just adopting @apgoetz's new API to replace this one (https://github.com/stm32-rs/stm32l0xx-hal/pull/105#pullrequestreview-465789802).

But if you can come up with a fix for the current API, that would of course be great.

PR #127, opened 2020-09-03
Detect rollover using positions rather than DMA flags

electroCutie:
I suspect that there is a timing issue with reading the flag then
clearing it in a non-atomic way. We can still use the flag to detect
overruns because it will never give a false positive, but it CAN give a
false negative it seems

Storing the position that we last witnessed the DMA as being at and
performing a wrap when that position is less than our current positon
gives a reliable detection and no fale overruns

We still examining the complete flag and the positions of the current
DMA pointer and our previous one to detect overruns since the fla will
never give a false positive

This is a fix for  #104

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -175,6 +175,7 @@ impl Adc<Ready> {
             ptr: buffer.as_ptr(),
             len: buffer.len() as u16,
             pos: 0,
+            dma_pos: 0,
 
             r_gt_w: false,
         };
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @electroCutie!

I'll need to take a close look and get my gear out for some testing, but I won't be able to get to that this week. I'll try to do it some time next week.

hannobraun:
> Anything I can check while I'm setup for it?

I don't have anything specific in mind, I just wanted to poke around a bit. I'm sure I'll come up with a more detailed testing plan, as I think more closely about the changes you made :-)

hannobraun (approved):
Sorry, it took me a bit longer to get to this than I hoped.

The changes look great! This is exactly the minimally invasive fix that I hoped for, but couldn't find myself. Good job, @electroCutie!

PR #129, opened 2020-10-03
Maximum frequency for stm32l0 increased to 32MHz

### LINKED ISSUE #110: PLL clock is incorrectly limited to 24MHz on l0x3 devices
PLL clock is limited here https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/rcc.rs#L307 to 24MHz, but l073 supports up to 32MHz.

akashihi:
Fixes #110

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -304,7 +304,7 @@ impl RccExt for RCC {
                     PLLDiv::Div3 => freq / 3,
                     PLLDiv::Div4 => freq / 4,
                 };
-                assert!(freq <= 24.mhz().0);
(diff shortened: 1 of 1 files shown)


akashihi:
@dbrgn May i ask you label the pr as 'hacktoberfest-accepted'? Looks like they have stricter rules this year.

dbrgn:
@akashihi done! As far as I understand the rules, it will count if it's either merged or labelled. Since I don't want to merge myself without someone else looking at it, I added the label.

arkorobotics:
Good catch! LGTM

PR #131, opened 2020-10-15
Adds support of LSE clocking to LPUART

### LINKED ISSUE #113: LPUART1 won't work with LSE as clock source
Bottomline: baudrate not set correctly.

I've been trying to print a simple "Hello world" with serial, using LPUART1. I wanted to make sure that LSE was being used as the clock source, to make it truly low power, and I ran into a problem. I used the serial.rs example from the repo, changing the USART to LPUART, and before the clock setup I added the following code: 

```
dp.RCC.csr.modify(|_,w| w.lseon().on()); //enable LSE
    while dp.RCC.csr.read().lserdy().is_not_ready() {    
        // wait until LSE is ready
    }

dp.RCC.ccipr.modify(|_,w| w.lpuart1sel().lse()); //set LSE as LPUART1 clock source
```
If I do the exact same thing, but instead of .lse() use .system() or .ahb(), it works fine. With .lse() it just won't print anything. So I started checking registers in a GDB session and found out the following:
LSEON gets set correctly, and LSERDY is set as expected. Then LPUART1SEL is correctly set to 11: LSE selected as clock.

What doesn't look right is the value of LPUART_BBR, and I think I understand, why: the function that calculates the LPUARTDIV takes fck from the system clock, and that's why it all works if .system() is used. In my code the system clock is MSI running at 1.048 MHz. But to correctly set the baudrate I need the actual LPUART clock, so in this case 32kHz. 

I'm just pointing this out, I wouldn't really know how to fix it: I guess some match or if/else block based on the LSE/other choice. This is how CubeMX does it:

```

      case UART_CLOCKSOURCE_SYSCLK:
        lpuart_ker_ck_pres = HAL_RCC_GetSysClockFreq();
        break;
      case UART_CLOCKSOURCE_LSE:
        lpuart_ker_ck_pres = (uint32_t)LSE_VALUE;
        break;
```

I hope this will help someone who knows what their doing :) to improve this, meanwhile I'm trying to figure out how to go around it.

akashihi:
Fixes #113 

Based on PR #130

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -735,7 +735,16 @@ usart! {
 }
 
 impl Serial<LPUART1> {
-    pub fn use_lse(&self, rcc: &mut Rcc, _: &LSE) {
+    /// Switches LPUART1 clock course to LSE
+    ///
+    /// Consumes LSE token, to get guarantee that
+    /// LSE clocks are configured.
+    ///
+    /// At the moment configured baudrate is ignored
+    /// and LPUART1 is forced to 9600 when clocked by LSE,
+    /// assuming that LSE is 32768
+    /// (and it must be so according to RM).
+    pub fn use_lse(&mut self, rcc: &mut Rcc, _: &LSE) {
         //Disable transmitter
         self.usart.cr1.modify(|_, w| w.te().disabled());
         while self.usart.isr.read().tc().bit_is_clear() {}
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -8,7 +8,7 @@ use crate::gpio::{AltMode, PinMode};
 use crate::hal;
 use crate::hal::prelude::*;
 pub use crate::pac::{LPUART1, USART1, USART2, USART4, USART5};
-use crate::rcc::Rcc;
+use crate::rcc::{Rcc, LSE};
 
 #[cfg(any(feature = "stm32l0x2", feature = "stm32l0x3"))]
 use core::{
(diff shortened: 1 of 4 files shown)


akashihi:
@dbrgn @arkorobotics May i ask you to review this one and #130? 

Thanks in advance.

hannobraun:
Hey @akashihi, thank you for the PR, and sorry again for the long wait! I've merged #130 and there are conflicts now. Can you please rebase this PR?

akashihi:
@hannobraun Conflicts fixed

src/serial.rs
```diff
@@ -734,6 +734,32 @@ usart! {
     USART5: (usart5, apb1enr, usart5en, apb1_clk, Serial5Ext),
 }
 
+impl Serial<LPUART1> {
+    pub fn use_lse(&self, rcc: &mut Rcc, _: &LSE) {
```
hannobraun, line 738: Is the `&self` a deliberate choice? I think it could be very surprising, if someone held a shared reference to `Serial` and somehow the configuration suddenly changed (especially since that might change baud rate with the current implementation). Unless there's a good reason not to, I'd prefer `&mut self`.

src/serial.rs
```diff
@@ -734,6 +734,32 @@ usart! {
...
+
+        //Disable LPUART1
+        self.usart.cr1.modify(|_, w| w.ue().disabled());
+
+        //Reconfigure LPUART to use LSE
+        rcc.rb.ccipr.modify(|_, w| w.lpuart1sel().lse());
+
+        //Recalculate baudrate
+        //TODO requested baudrate value from the config should be stored somehow and used here
+        //but at the moment hardcoded 9600 will be used
```
hannobraun, line 751: I think this is fine for now, but it should be mentioned in the documentation of the method.

hannobraun (changes requested):
Thanks again, @akashihi! I reviewed the commits and found some minor issues. Let me know what you think.

I don't have the time to test this, but if it works for you, that's good enough.

src/serial.rs
akashihi: @hannobraun Good catch :)

src/serial.rs
akashihi: Looks like method needs documentation too  👍

hannobraun (approved):

hannobraun:
Great, thank you @akashihi!

PR #132, opened 2020-11-03
remove depreciated features from #cfg guards

berkowski:
The 'stm32l052', 'stm32l072', and 'stm32l082' features are marked
depreciated and not intended to be used by consumers but they are
frequently used to enable specific features within the API.

This commit aliases the features to family/category pairs and
changes the #cfg guards to match

This WILL create issues when a specific package does not provide
all the pins, as the old features are really a family/category/package
triplet.  As this change is permissive and should only include non-physical 
pins (as opposed to exclude pins on the package) my opinion is that should
be fixed in follow-up PR's


hannobraun:
Thank you for the PR, @berkowski! Sorry for the long wait.

This PR has conflicts with the master branch now and needs a rebase. Otherwise this PR looks good, although I haven't checked against the various data sheets. Maybe @dbrgn wants to take a look.

Cargo.toml
```diff
@@ -158,29 +161,6 @@ mcu-STM32L051R8Hx = ["io-STM32L051", "stm32l0x1", "tfbga64"]
...
-mcu-STM32L053C6Tx = ["io-STM32L051", "stm32l0x3", "lqfp48"]
-mcu-STM32L053C8Tx = ["io-STM32L051", "stm32l0x3", "lqfp48"]
-mcu-STM32L053R6Hx = ["io-STM32L051", "stm32l0x3", "tfbga64"]
-mcu-STM32L053R6Tx = ["io-STM32L051", "stm32l0x3", "lqfp64"]
-mcu-STM32L053R8Hx = ["io-STM32L051", "stm32l0x3", "tfbga64"]
-mcu-STM32L053R8Tx = ["io-STM32L051", "stm32l0x3", "lqfp64"]
-mcu-STM32L062K8Tx = ["io-STM32L051", "stm32l0x2", "lqfp32"]
-mcu-STM32L062K8Ux = ["io-STM32L051", "stm32l0x2", "ufqfpn32"]
-mcu-STM32L063C8Tx = ["io-STM32L051", "stm32l0x3", "lqfp48"]
-mcu-STM32L063R8Tx = ["io-STM32L051", "stm32l0x3", "lqfp64"]
```
dbrgn, line 183: Regarding the re-grouping: Note that this code is generated (with https://github.com/dbrgn/cube-parse/), so I'd suggest not to group them by hand (or alternatively, extend the generator to support these comments).

src/adc.rs
```diff
@@ -676,15 +676,24 @@ adc_pins! {
     Channel9: (gpiob::PB1<Analog>, 9u8),
 }
 
-#[cfg(all(feature = "stm32l052", any(feature = "lqfp64", feature = "tfbga64",),))]
+#[cfg(all(
+    all(
+        any(feature = "io-STM32L051", feature = "io-STM32L071"),
+        any(feature = "stm32l0x2", feature = "stm32l0x3"),
```
dbrgn, line 682: Do we still need these two features, or can we remove them from the cfg filter? (In other words, does `io-*` handle everything already?)

berkowski:
I've since moved on to using my own HAL layer for the stm32l071x that's better suited to my needs.

dbrgn:
Thanks for letting us know. In that case, I'll close this PR!

PR #136, opened 2020-12-08
Simplify flash.c

dbrgn:
The manual for-loop looked a bit funny :)

(A `memcpy` would probably work as well, but is less easy to read and also results in a larger binary.)

### DIFF
--- a/flash-code/flash.c
+++ b/flash-code/flash.c
@@ -5,9 +5,7 @@
 // function is necessary.
 __attribute__((section(".data")))
 void write_half_page(uint32_t *address, uint32_t *words) {
-    uint32_t i = 0;
-    while (i < 16) {
+    for (uint32_t i = 0; i < 16; i++) {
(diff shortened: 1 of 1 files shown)


hannobraun (commented):
Thanks!

Did you call `compile.sh`? If it generates different code, we should update the binary also.

dbrgn:
On `master`:

    $ ./compile.sh && sha1sum libflash.a
    0f720d2f86e55e2fe2e05bc639c18eef595a45fa  libflash.a

On this branch:

    $ ./compile.sh && sha1sum libflash.a
    0f720d2f86e55e2fe2e05bc639c18eef595a45fa  libflash.a

...so I didn't commit the rebuilt binary, especially considering that the checked-in binary is 896 bytes, while the one my compiler generates is 900 bytes. Probably because a different gcc version was used.

hannobraun:
Makes sense, thank you.

PR #140, opened 2021-01-05
Flash: Support writing single bytes to EEPROM

dbrgn:
EEPROM supports single-byte writes.

image

This also changes the pointer type from `u32` to `usize.

### DIFF
--- a/src/flash.rs
+++ b/src/flash.rs
@@ -38,8 +38,8 @@ pub const EEPROM_SIZE: usize = 512;
 pub const EEPROM_SIZE: usize = 128;
 
 // EEPROM start addresses
-const EEPROM_START_BANK1: usize = 0x0808_0000;
-const EEPROM_START_BANK2: usize = 0x0808_0C00;
+pub const EEPROM_START_BANK1: usize = 0x0808_0000;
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Looks good to me. I haven't tested, but since it seems to work for you, that's good enough for me.

I'm not too sure about changing the pointer type to `usize`. As far as I can see, that doesn't make any difference, except to make the code less explicit. I don't think it's a big deal though.

I'm leaving this open in case we want to talk about the `usize` change some more, but as far as I'm concerned, this can be merged.

dbrgn:
> As far as I can see, that doesn't make any difference, except to make the code less explicit.

For me, it's *more* explicit. The type is now "unsigned integer the size of pointer" (which is fitting since it stores pointers), while u32 is just "a 32 bit unsigned integer". For me it's mostly about the semantics, not about the actual size.

hannobraun:
> For me, it's _more_ explicit. The type is now "unsigned integer the size of pointer" (which is fitting since it stores pointers), while u32 is just "a 32 bit unsigned integer". For me it's mostly about the semantics, not about the actual size.

Yeah, that makes sense. Convinced.

PR #142, opened 2021-01-28
fix Pin::into_pull_down_input

rkreis:
There was a typo there... Configuring a Pin as pull-down would actually set it as a floating input.


dbrgn (approved):
LGTM, probably a copy-paste error!

PR #143, opened 2021-02-07
Add MCO support

jamwaffles:
I've added MCO (Microcontroller Clock Output) support in a hopefully sensible and typesafe way.

I'm reasonably happy with the impl, but will happily take any improvements. I tested the example with my L053 Nucleo using HSI16 as the clock reference. I can get PA8 and/or PA9 to show a nice steady 2MHz, so the code seems to be working as expected.

MCO consumes the output pin, which I think the API reflects. I'm also assuming MCO is available on all variants, so nothing is feature gated in this PR.

### DIFF
--- a/examples/mco.rs
+++ b/examples/mco.rs
@@ -1,0 +1,40 @@
+//! Configure the MCO (Microcontroller Clock Output) to give a 2MHz signal on PA8 and PA9, sourced
+//! from the internal HSI16 16MHz oscillator.
+
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_halt;
+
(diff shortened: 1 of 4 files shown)


jamwaffles:
I've rebased as #144 is merged. Would appreciate another look from @dbrgn or @hannobraun!

Aside: Nightly has started passing in Travis :thinking:

hannobraun:
Thank you, @jamwaffles!

I'm not familiar with this feature, but the code looks good to me. And it's a new feature anyway, so unlikely to break existing use cases. Merging.

jamwaffles:
Awesome! Thank you :blush:

PR #144, opened 2021-02-09
Add QEI support for LPTIM

jamwaffles:
As per the title, adds support for turning on encoder mode for LPTIM. e.g.

```rust
let mut pwr = PWR::new(dp.PWR, &mut rcc);
    
let lptim = LpTimer::init_encoder(
    dp.LPTIM,
    &mut pwr,
    &mut rcc,
    ClockSrc::Hsi16,
    (gpiob.pb5, gpiob.pb7),
    // Roll over at 4096
    4096,
);

defmt::info!("Count {:?}", lptim.count());
```

### DIFF
--- a/src/lptim.rs
+++ b/src/lptim.rs
@@ -1,5 +1,6 @@
 //! Low-Power Timer (LPTIM) support.
 
+use crate::gpio::{self, gpiob};
 use crate::hal;
 use crate::pac::LPTIM;
 use crate::pwr::PWR;
@@ -20,14 +21,19 @@ pub enum OneShot {}
 /// Low-Power Timer counting in periodic mode.
 pub enum Periodic {}
 
+/// Low-Power Timer in encoder mode.
+pub enum Encoder {}
+
 impl sealed::Sealed for OneShot {}
 impl sealed::Sealed for Periodic {}
+impl sealed::Sealed for Encoder {}
 
 /// Marker trait for counter directions.
 pub trait CountMode: sealed::Sealed {}
 
 impl CountMode for OneShot {}
 impl CountMode for Periodic {}
+impl CountMode for Encoder {}
 
 /// Clock source selection for the Low-Power Timer `LPTIM`.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


jamwaffles:
This was tested on an L053 Nucleo with a lil hand encoder module from Aliexpress.

jamwaffles:
I just realised it's not possible to enable interrupts with this API. Converting to draft until that's fixed...

jamwaffles:
The API looks like this now:

```rust
let mut encoder = LpTimer::init_encoder(
    LPTIM,
    &mut pwr,
    &mut rcc,
    // Use low speed internal oscillator
    lptim::ClockSrc::Lsi,
    (gpiob.pb5, gpiob.pb7),
);

encoder.enable_interrupts(lptim::Interrupts {
    enc_dir_up: true,
    ..lptim::Interrupts::default()
});

// Set ARR to 4096
encoder.enable(4096);
```

I don't think it's very ergonomic, but at least it works... it also reuses as much of the existing API as possible.

Very open to suggestions on how to make this nicer to use!

src/lptim.rs
```diff
@@ -105,6 +111,81 @@ impl LpTimer<OneShot> {
...
+        // actually enabled."
+        // The slowest LPTIM clock source is LSE at 32768 Hz, the fastest CPU clock is ~80 MHz. At
+        // these conditions, one cycle of the LPTIM clock takes 2500 CPU cycles, so sleep for 5000.
+        cortex_m::asm::delay(5000);
+
+        // ARR can only be changed while the timer is *en*abled
+        self.lptim.arr.write(|w| w.arr().bits(arr));
+    }
+
+    pub fn enable(&mut self, arr: u16) {
```
dbrgn, line 169: Is there a way to disable the timer again?

dbrgn, line 169: To me it's not obvious what `arr` contains, but that may be because I'm not very familiar with QEI. However, since it's a public method, it probably wouldn't hurt to add a documentation string to this function with a quick explanation.

dbrgn (commented):
I'm not really familiar with QEI (haven't used it myself so far), but I assume since @jamwaffles tested this, we can merge the PR once the comments are addressed. The code looks clean.

Maybe some other maintainers have experience with QEI?

hannobraun (approved):
I'm not familiar with QEI either, but I can't see anything wrong here. It works for @jamwaffles, so let's merge it!

jamwaffles:
Review comments addressed - I added a more descriptive comment for `arr`, although I left it vague as I'm unsure what it wraps around to. The value in `CMP` maybe?

I also added a `disable` method as suggested.

hannobraun (approved):
Thank you, @jamwaffles, looks good to me. Not merging, to give @dbrgn a chance to take another look.

threema-danilo:
Perfect, thanks @jamwaffles!

dbrgn (approved):
Perfect, thanks @jamwaffles!

jamwaffles:
Thanks! One less patch to keep in my fork :P

PR #145, opened 2021-02-19
Add encoder input support for TIM2 and TIM21

jamwaffles:
Adds encoder (QEI) input support to TIM2 and TIM21. I've tested the interrupt handling with an RTIC 0.6 project, and have included a short non-RTIC example to get users off the ground.

This leaves a few pin combinations and a few timers without an implementation currently, as I have no way to test them (or couldn't get them working) - I'll leave that as a job for other people who need it :)

### DIFF
--- a/examples/timer_encoder.rs
+++ b/examples/timer_encoder.rs
@@ -1,0 +1,39 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_halt;
+
+use cortex_m_rt::entry;
+use stm32l0xx_hal::{
+    encoder::{Mode, Status},
+    pac,
+    prelude::*,
+    rcc::Config,
+};
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Configure the clock.
+    let mut rcc = dp.RCC.freeze(Config::hsi16());
+
+    let gpiob = dp.GPIOB.split(&mut rcc);
+
+    // Create an encoder instance that counts between 0 and 64 using inputs on PB13 and PB14.
+    let mut encoder = dp
+        .TIM21
+        .encoder((gpiob.pb13, gpiob.pb14), Mode::Qei, 64, &mut rcc);
+
+    loop {
+        #[allow(unused)]
+        let Status {
+            count,
+            did_overflow,
+            direction,
+        } = encoder.status();
+
+        // Use encoder state here
+    }
+}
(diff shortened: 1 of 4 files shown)


hannobraun (approved):
As usual, I'm not very familiar with the subject matter, but can't find anything wrong. Since it works for @jamwaffles, I'm going to merge it.

Two notes:

- I'm not a fan of the `EncoderExt` trait (and those kind of extension traits in general). Just seems like unnecessary complexity where a simple constructor will do.
- I don't like that most of the implementation lives in a macro. I much prefer to abstract over peripheral instances with an `Instance` trait (whose implementations are macro-generated) and build as much of the code as possible on top of that `Instance` trait, outside of the macro. See [pwm.rs](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/pwm.rs), for example.

But both of these techniques are widespread in the stm32-rs world, and I don't see the need to fight against windmills right now :smile:. Feel free to send a follow-up PR if you agree and want to work on this, but if not, this is fine too.

jamwaffles:
Thanks for the review and the feedback!

> I'm not a fan of the EncoderExt trait (and those kind of extension traits in general). Just seems like unnecessary complexity where a simple constructor will do.

That's fair. I mirrored the API in `timer.rs` so one can do `dp.TIMx.encoder` which seems to be somewhat common, at least in RTIC contexts.

> I don't like that most of the implementation lives in a macro. I much prefer to abstract over peripheral instances with an Instance trait (whose implementations are macro-generated) and build as much of the code as possible on top of that Instance trait, outside of the macro. See pwm.rs, for example.

Also fair! I'd prefer to do it without macros for the record. I took a brief look at doing it without them and gave up pretty quickly. I'll have another go and open a new PR if I get a clean solution together.

hannobraun:
> That's fair. I mirrored the API in `timer.rs` so one can do `dp.TIMx.encoder` which seems to be somewhat common, at least in RTIC contexts.

Yes, it's definitely common, but I think that's mostly cargo cult. At least I haven't heard anyone argue that it should be that way. And there's definitely been a broad movement away from that approach in various HALs.


> Also fair! I'd prefer to do it without macros for the record. I took a brief look at doing it without them and gave up pretty quickly. I'll have another go and open a new PR if I get a clean solution together.

Yeah, we usually need macros to abstract over the different peripheral instances. Things get messy quick without that. But usually it's possible to minimize the code within the macro using the `trait Instance` approach. Sometimes it's not easy, unfortunately, if there are many subtle differences between peripheral instances.

But as I said in the review, even though a follow-up PR is definitely welcome, I'm fine with this as-is.

PR #148, opened 2021-03-04
Add DMA support to SPI API

michaelbeaumont:
It's not completely clear from the reference manual if we need to deal with `SPI_SR`.`OVR` or `SPI_SR`.`RXNE` here or if this can be made more robust.

### DIFF
--- a/examples/spi_dma.rs
+++ b/examples/spi_dma.rs
@@ -1,0 +1,89 @@
+#![no_main]
+#![no_std]
+
+extern crate panic_halt;
+
+use core::pin::Pin;
+
+use cortex_m::interrupt;
+use cortex_m_rt::entry;
+use stm32l0xx_hal::{
+    dma::{self, DMA},
+    pac::{self, Interrupt, NVIC},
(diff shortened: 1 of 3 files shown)


hannobraun (commented):
Thank you @michaelbeaumont! I don't have the time to test, but the code looks good to me.

Did you successfully test this? Adding an example to the repository would be ideal, but if one isn't easily available, I'm happy to accept your word that is works you for, and merge this as-is :smiley: 

> It's not completely clear from the reference manual if we need to deal with `SPI_SR`.`OVR` or `SPI_SR`.`RXNE` here or if this can be made more robust.

I don't know either, but if this works for you, I'm happy to merge it. Improvements can always be made later, once someone with the necessary expertise has the time/motivation to look into it.

michaelbeaumont:
It works for me (I've only tested the `write` functionality to be honest). I'll come up with an example!

hannobraun (approved):
Looks good to me. Than you, @michaelbeaumont!

PR #149, opened 2021-03-07
v0.6.3 - Timer improvements

rkreis:
Is it time for a new release after 10 months? :)


hannobraun:
I agree, it's probably time. We might have had some breaking changes though since the last release? Unless someone is willing to do an audit of the merged pull request to make sure, it's probably better to release as 0.7.0, to be safe.

I've made a note to take care of it and will try to get to it within the next few days. Looks like we don't even have a changelog here, which makes things easy :smile:

dbrgn:
A changelog would be nice to have though, for the future.

PR #152, opened 2021-03-11
Add `Timer::new` to create timer without starting

michaelbeaumont:
Useful for initializing a `Timer` to call `start()` on later.

I suppose `timer()` creates it _and_ starts it as a sort of convenience, but it's a little silly having to pass a dummy timeout. More seriously, calling `timer()` with a dummy timeout but without `listen()` and later calling `listen()` and `start()` can cause a subtle issue, I believe. If the dummy timeout has already overflowed, the interrupt is already pending and will immediately fire, before the new `start` has expired (necessitating a `clear_irq()` call in addition to `listen` and `start`).

This PR doesn't fix that issue per se, it only avoids it for creation but I think the expected behavior for `start()` would be that it also calls `clear_irq`, WDYT?

This is a breaking change for the `GeneralPurposeTimer` trait but it feels reasonable, though we could also add an additional trait to hold `enable`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@ forget to update the links at the bottom of the changelog as well.-->
 
 ### Additions
 
+- Add `Timer::new` to create timer without starting #152
+
 ### Breaking Changes
+
+- Add `enable` to `GeneralPurposeTimer`
 
 ### Non-Breaking Changes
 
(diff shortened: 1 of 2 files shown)


dbrgn:
Thanks! This looks reasonable to me, but could you add documentation to the trait and its methods?

(Also, we should really set up a CHANGELOG. I think I'll open a PR with a suggestion.)

michaelbeaumont:
Done, let me know if that's sufficient. I think the trait is sort of only internally used at the moment, so there's not much to say about them.

dbrgn:
Thanks!

PR #154, opened 2021-03-11
Add a CHANGELOG

dbrgn:
Format is loosely based on keepachangelog.com, the format that a lot of
other STM32 HAL crates use as well.

The main issues with single-file changelogs is that they cause a lot of
conflict potential. One solution is writing changelog entries to
separate files, and merging those before a release. However, that
requires additional tooling and is a bit cumbersome.

Instead, I added `CHANGELOG.md merge=union` to `.gitattributes`. I'm not
sure if it always works ideally, but at least it should reduce the
conflict potential a lot. And in the case something breaks, we can
always fix it manually.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.md merge=union
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,133 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/).
+
+
+
+## [Unreleased]
+
+<!-- When making a PR, please update this section. Note: This document should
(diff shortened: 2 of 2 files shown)


dbrgn:
PS: I also tagged all releases since v0.5.0!

hannobraun:
Thank you, @dbrgn, great work!

I'm a bit skeptical of requiring contributors to add changelog entries. That puts a higher burden on them, which might be unreasonable to expect in the case of small drive-by contributions. It's also another thing to remember on the maintainer side, and I suspect it will lead to some pull request piling up, if we're strict about it.

Still, this PR is definitely an improvement, and we can work out the kinks as we go. Maybe we could just add a tag for PRs that are merged without changelog entries, and have someone go through them before the next release (preferably whoever is asking for a release to be made :smile:).

PR #155, opened 2021-03-15
Allow skipping linker script

lulf:
By counting the features before the "disable-linker-script" feature is checked, it will fail even if this flag was set.


hannobraun (changes requested):
Thank you, @lulf! Can you please run `cargo fmt`? CI is failing due to formatting.

Looks good otherwise!

PR #156, opened 2021-03-18
Fixed bad op for RTC wakeup

allexoll:
`^` was probably intended as "power" but is XOR in this context, limiting delay to a maximum of 2 XOR 17, or ..=19

### DIFF
--- a/src/rtc.rs
+++ b/src/rtc.rs
@@ -471,7 +471,7 @@ impl timer::CountDown for WakeupTimer<'_> {
         T: Into<Self::Time>,
     {
         let delay = delay.into();
-        assert!((1..=2 ^ 17).contains(&delay));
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @allexoll, good catch!

One note: I think the method documentation is somewhat wrong. Under "Panics", it states `<= 2^17`, which doesn't match the new implementation, or the rest of the method's documentation. It should probably be `< 2^17`.

I'm merging this PR now, since it's a strict improvement (the documentation is not more wrong than it was before), and am going to open an issue to track the documentation problem. A follow-up PR that fixes the documentation issue would be very welcome!

allexoll:
done in [PR#157](https://github.com/stm32-rs/stm32l0xx-hal/pull/157)

PR #157, opened 2021-03-19
fixed documentation error for RTC wakeup

allexoll:
fixed documentation as mentioned in [PR#156](https://github.com/stm32-rs/stm32l0xx-hal/pull/156)

PR #158, opened 2021-03-19
Fix RTC documentation issue

hannobraun:
Follow-up to https://github.com/stm32-rs/stm32l0xx-hal/pull/156. Turns out I can make that change right here on GitHub, which is about as much work as opening an issue would be :smile: 

cc @allexoll


hannobraun:
@allexoll was quicker than me. Closing in favor of #157.

PR #162, opened 2021-04-07
added example temperature.

pdgilbert:
This example builds but has not been tested on hardware.


hannobraun:
Thanks for the pull request, @pdgilbert, looks good to me!

The Clippy build fails but the code it fails on hasn't been touched by this PR. Merging.

PR #165, opened 2021-04-28
added support for Device electronic signature

allexoll:
this was talked briefely in #120, i think @hannobraun 's idea is also the easiest to add.

code was indeed copied from [stm-32-device-signature](https://github.com/stm32-rs/stm32-device-signature/blob/master/src/lib.rs), credit to @Disasm

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,6 +57,7 @@ pub mod rtc;
     feature = "io-STM32L071",
 ))]
 pub mod serial;
+pub mod signature;
 pub mod spi;
 pub mod syscfg;
 pub mod time;
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -3,30 +3,37 @@
 //! The STM32L0 contains a few read-only registers with factory programming
 //! data that are written during production.
 
-
 use cortex_m::interrupt;
 
 mod pointers {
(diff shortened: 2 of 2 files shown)


Disasm:
Note that `device_id()` code for L0 is incorrect: https://github.com/stm32-rs/stm32-device-signature/issues/9

allexoll:
Oh thanks for the heads up, i'll try to fix that

allexoll:
according to the reference manual, it is at least coherent between x0,x1,x2,x3 family. all have the same memory layout. 

I think it would be better to have the user give an array to fill then. or other possibility is doing something similar to [stm32f7xx](https://github.com/stm32-rs/stm32f7xx-hal/blob/f1e36d9a6bdc9d7d01d35327bbb276eca9291dc4/src/signature.rs)

i'll look into it tomorrow.

allexoll:
this is my proposal for the UID interface. it returns the 96bit uid as a contiguous buffer, since i think it is the way that most will use this API. remarks are welcome, sorry for the multiple commits.

hannobraun (approved):
Thank you, @allexoll, looks good to me!

No time really to test or double-check against the reference manual, as usual, but happy to merge if it works for you.

Issue #166, opened 2021-04-30
update PAC from 0.10.0 to 0.13.0


allexoll:
there has been updates to the pac. 

some fields were missing (namely in the CRC peripheral, and i'm porting the F4 CRC driver to this family, but some fields aren't present in the 0.10.0 pac. 

trying to change the pac version only seems to break 2 calls to unsafe functions in the DMA driver, so that seems fixable.

What is your opinion on this?

```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> /Users/allexoll/Desktop/stm32l0xx-hal/src/dma.rs:344:55
    |
344 |                       handle.dma.$chfield.par.write(|w| w.pa().bits(address));
    |                                                         ^^^^^^^^^^^^^^^^^^^^ call to unsafe function

error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> /Users/allexoll/Desktop/stm32l0xx-hal/src/dma.rs:351:55
    |
351 |                       handle.dma.$chfield.mar.write(|w| w.ma().bits(address));
    |                                                         ^^^^^^^^^^^^^^^^^^^^ call to unsafe function
```

hannobraun:
Replied in #167.

allexoll:
has been adressed in #168

PR #167, opened 2021-04-30
Update cargo.toml to pac v0.13

allexoll:
followong #166,  updated the pac and dependencies. fixed some issues with DMA access needing to be unsafe since pac update

Remarks welcome

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -341,14 +341,14 @@ macro_rules! impl_channel {
                     handle:  &mut Handle,
                     address: u32,
                 ) {
-                    handle.dma.$chfield.par.write(|w| w.pa().bits(address));
+                    handle.dma.$chfield.par.write(|w| unsafe{w.pa().bits(address)});
                 }
 
                 fn set_memory_address(&self,
                     handle:  &mut Handle,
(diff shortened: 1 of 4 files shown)


src/dma.rs
```diff
@@ -341,14 +341,14 @@ macro_rules! impl_channel {
                     handle:  &mut Handle,
                     address: u32,
                 ) {
-                    handle.dma.$chfield.par.write(|w| w.pa().bits(address));
+                    handle.dma.$chfield.par.write(|w| unsafe{w.pa().bits(address)});
```
hannobraun, line 344: I suspect this (and the corresponding line below) is perfectly fine, but I'd prefer if you double-checked against the reference documentation, to see which values are actually allowed in this field, and add a comment to explain why this use of `unsafe` is sound (e.g. "This is sound, as PA accepts all possible values of `u32`"; don't know if that's the case, just giving you an example of the kind of comment I'm looking for).

hannobraun (commented):
Thank you, @allexoll! I left one comment. Looks good otherwise, but could probably use a rebase, now that #165 is merged.

PR #168, opened 2021-05-01
fixed update and added comments to unsafe dma access

allexoll:
sorry for the double PR, i accidentally deleted my #167 branch... 

anyway, here we go, i added the comments, your remarks were spot on.


hannobraun (approved):
Thank you, @allexoll, looks good!

PR #169, opened 2021-05-03
adding support for crc engine, shamelessly taken from stm32l4xx-hal.

allexoll:
I also added an example for CRC32 (Ethernet) since User manual is unclear. It says that the default CRC config is for CRC32 (Ethernet), but only the polynomial is correct, you still need to rev in and rev out bits, as well as xor the final result.

There still is an example for custom polynomial in the header doc of the CRC file, and i think both of them combined show how to use the CRC peripheral. no changes were needed for this to work on the L0xx family.

remarks welcome

### DIFF
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -1,0 +1,210 @@
+//! CRC calculation unit
+//!
+//! This unit is modeled after the [corresponding unit in the stm32l4xx-hal](https://github.com/stm32-rs/stm32l4xx-hal/blob/master/src/crc.rs).
+//!
+//! Usage example:
+//! ```
+//! let crc = dp.CRC.constrain(&mut rcc.ahb1);
+//!
+//! // Lets use the CRC-16-CCITT polynomial
+//! let mut crc = crc.polynomial(crc::Polynomial::L16(0x1021)).freeze();
+//!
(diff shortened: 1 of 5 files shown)


src/crc.rs
```diff
@@ -0,0 +1,210 @@
...
+                w.rev_out().clear_bit()
+            }
+        });
+
+        Crc {}
+    }
+}
+
+/// Constrained CRC peripheral.
+pub struct Crc {}
```
hannobraun, line 145: I think it would be nicer to store `CRC` in here, then we wouldn't need the `unsafe` blocks in most of the method below. But again, good enough.

hannobraun (approved):
Thanks again, @allexoll! I left a few comments because I love nitpicking, but this is fine. Merging!

allexoll:
thanks for the remarks, i'll try to implement them when/if i have the time. i think in this case since there was no support for the CRC peripheral, some support is better than no support, so this was a quick job.

hannobraun:
> i think in this case since there was no support for the CRC peripheral, some support is better than no support, so this was a quick job.

Fully agreed, that's why I merged it. Further improvements are welcome, of course, but I'd rather have this module in a style I don't prefer than not have it at all (or have a pull request sit open forever).

PR #173, opened 2021-06-10
Make flash and ram size depend on the mcu feature

### LINKED ISSUE #170: stm32l051k8u6 has 64kb of flash but is configured with 16kb
Hi!

Unless I'm missing something, I think that the memory descriptions loaded are not always correct? I'm using for example a `stm32l051k8u6` and `stm32l071k8u6`. According to the `Cargo.toml` of this hal it should activate the following features:
```toml
mcu-STM32L051K8Ux = ["stm32l0x1", "ufqfpn32", "io-STM32L051", "eeprom-2048"]

# Or

mcu-STM32L071K8Ux = ["stm32l0x1", "ufqfpn32", "io-STM32L071", "eeprom-3072"]
```

So they both fall under the `stm32l0x1` group and that loads `memory_l0x1.x` according to the build.rs file. This defines that there is `16kb` of flash. However on both of those MCUs the flash should be `64kb` according to the datasheet.
```
MEMORY
{     
    FLASH : ORIGIN = 0x08000000, LENGTH = 16K     
    RAM : ORIGIN = 0x20000000, LENGTH = 2K
}
```

Screenshot from 2021-06-10 10-35-23
Screenshot from 2021-06-10 10-35-06

This causes some linking issues because the firmware is "too big" to fit in 16K although it works fine when I select the `stm32l0x3` feature.

1. Can I override the memory configuration in my code without changing this hal crate as a temporary workaround?
2. Can we have more granularity in the flash memory configurations?

azerupi:
This fixes #170 by adding features for flash and RAM sizes and enabling the appropriate one for each mcu feature. It uses the new output from dbrgn/cube-parse#5 

I'm not sure why I have more MCUs in the list, maybe a more recent version of CubeMX?
I tested it out locally, it seems to work.

Also I think this is a **breaking change**. Because it might break code that does not select a `mcu-*` feature but selects the other features manually. In that case no memory.x file will be loaded and the build will fail.


hannobraun (commented):
Looks good to me, although I didn't test or check against data sheets. I'd like to hear @dbrgn's opinion, as he's the resident CubeMX expert.

azerupi:
No problem, it's better to take the time to review and test this :slightly_smiling_face: 

I just realized I didn't remove the memory_l0x*.x files, so I'll do that before it gets merged. 

On an unrelated note, I was wondering what config you guys where using for rust-analyzer? I've added the following `.vscode/settings.json` file, but I'm wondering if you have more elaborate configurations that work better?
```json
{
    "rust-analyzer.cargo.features": ["mcu-STM32L071K8Ux"],
}
```

dbrgn:
> Looks good to me, although I didn't test or check against data sheets. I'd like to hear @dbrgn's opinion, as he's the resident CubeMX expert.

🙂 I'll take a look as soon as I can.

hannobraun:
> On an unrelated note, I was wondering what config you guys where using for rust-analyzer? I've added the following `.vscode/settings.json` file, but I'm wondering if you have more elaborate configurations that work better?

It's actually been a while since I actively worked on this repository, and I don't have an active checkout. I do have `"rust-analyzer.checkOnSave.allTargets": false`, in addition to the `cargo.features`, in the configuration for another HAL. But I don't remember why I needed that.

dbrgn:
> I'm not sure why I have more MCUs in the list, maybe a more recent version of CubeMX?

I think so, yes! I updated the list with the old version of cube-parse in #178. Can you rebase against the current `master` branch and resolve the conflicts by simply overwriting the conflicting part with the generated features? That way, only the new `ram-*` and `flash-*` feature changes should remain.

In case you're unsure how to do this with git, don't hesitate to ask, I'm happy to help!

README.md
```diff
@@ -79,19 +79,10 @@ for example the `STM32L062K8Tx` uses the GPIO peripheral version named
 
 ## Flash memory
 
-Currently a default Flash and RAM size are configured depending on the chip's subfamily.
+Flash and RAM sizes are configured for each mcu in the feature list.
```
dbrgn, line 82: I'd slightly reword this:

```suggestion
Flash and RAM sizes are automatically configured by using the appropriate
`mcu-*` Cargo feature, as described further up.
```

README.md
```diff
@@ -79,19 +79,10 @@ for example the `STM32L062K8Tx` uses the GPIO peripheral version named
...
-| `stm32l0x2` | 192kB      | 20kB     |
-| `stm32l0x3` | 64kB       | 8kB      |
-
-However not all chips in the subfamily have the same flash and RAM size. Due to a known limitation
-to the scripts used to generate the features, the sizes might not correspond to the true sizes of 
-the MCU. You should double check with the datasheet, to make sure the flash and RAM sizes are correct.
-
-If they are not correct, you can override the `memory.x` of `stm32l0xx-hal` by providing your own.
+If you have an mcu that is not in the list or they are not correct, you can override 
+the `memory.x` of `stm32l0xx-hal` by providing your own.
```
dbrgn, line 85: Slight improvement (hopefully):

```suggestion
If you have an MCU that does not yet have a corresponding `mcu-*` feature, or
if the default configuration is incorrect, you can override the `memory.x` of
`stm32l0xx-hal` by providing your own:
```

build.rs
```diff
@@ -26,25 +26,59 @@ fn main() {
...
+            (16, cfg!(feature = "flash-16")),
+            (32, cfg!(feature = "flash-32")),
+            (64, cfg!(feature = "flash-64")),
+            (128, cfg!(feature = "flash-128")),
+            (192, cfg!(feature = "flash-192")),
+        ]
+        .iter()
+        .filter(|(_, f)| *f)
+        .map(|(f, _)| *f)
+        .collect();
```
dbrgn, line 40: Nice way to avoid an if-festival! 👍🏻

dbrgn (changes requested):
Two small suggestions for the README, otherwise this looks good to me! Thanks for implementing this change, I think it will be very useful to get started quickly.

azerupi:
Sorry for the noise, I think I forgot to add a file when rebasing. It should be good now.

dbrgn (approved):
Thanks! An intermediate commit adds git conflict markers (which are then removed again in the next commit), but we can solve that quite easily by squashing 🙂

Issue #172, opened 2021-06-10
Can't change PWM frequency after instanciating a Pwm struct


azerupi:
Hi,

I'm trying to change the frequency of a PWM pin dynamically. This is impossible with the currant API because the method to change the frequency is on the timer, but the timer contains all the pwm channels and when you 'associate' a pwm channel to a pin it is moved out of the timer struct. When you then try to configure the pwm frequency you get a 'borrow of partially moved value' error.

**Example:**
```rust
let mut pwm_timer = Timer::new(peripherals.TIM2, 100.hz(), &mut rcc);
let front_led_pwm_pin = pwm_timer.channel1.assign(gpioa.pa0);

// Some code

pwm_timer.set_frequency(10.hz(), &rcc);
```

Resulting in the following error:
```
error[E0382]: borrow of partially moved value: `pwm_timer`
   --> src/main.rs:104:5
    |
102 |     let front_led_pwm_pin = pwm_timer.channel1.assign(gpioa.pa0);
    |                                                ----------------- `pwm_timer.channel1` partially moved due to this method call
103 |
104 |     pwm_timer.set_frequency(10.hz(), &rcc);
    |     ^^^^^^^^^ value borrowed here after partial move
    |
note: this function takes ownership of the receiver `self`, which moves `pwm_timer.channel1`
```

## Workaround

I discussed this on the matrix channel a bit and I currently have two workarounds requiring unsafe:

1. Change the registers directly  
    ```rust
    unsafe {
        (*pac::TIM2::ptr()).psc.write(|w| w.bits(psc_val));
        (*pac::TIM2::ptr()).arr.write(|w| w.bits(arr_val));
    }
    ```
    This works but you loose the niceties provided by the `set_frequency` function (e.g. provide a frequency in Hz).
    
2. Make a raw pointer to the timer object to bypass Rust's ownership
    ```rust
    let pwm_timer_ptr: *mut Timer<TIM2> = &mut pwm_timer;
    // Some code
    unsafe { 
        (*pwm_timer_ptr).set_frequency(frequency.hz(), &context.rcc); 
    }
    ```
    This also compiles, however I'm not sure what the implications are. Is this "safe"? I guess it is as long as I don't try to associate the channel with another pin?

## HAL crate
Would it be possible to support this use case in the HAL library to avoid the need to use `unsafe`?

[...]

PR #174, opened 2021-06-11
Add set_frequency method on PWM channels

azerupi:
Following discussion in #172  this adds a method to PWM channels to change the underlying timer's frequency. This allows to dynamically change the PWM frequency when we have already associated a channel with a pin. Otherwise this required unsafe to work around a partial move error because the channel was moved out of the timer struct.

In the implementation I changed `clock_frequency` on the `Instance` to not take `&self`. This allows it to be used in the channels. But this is not required, we could access the RCC register directly to avoid making a breaking change?

Compared to the timer's implementation we don't stop and start the timer before changing the frequency. I tested it on my project and it seems to work flawlessly. I'm not sure why that was required in the timer's implementation? If it needs to be added, is there a way to reuse the timer's implementation without copy-pasting?

Let me know if anything can be improved! :slightly_smiling_face:

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -74,7 +74,7 @@ where
     /// frequency
     pub fn set_frequency(&mut self, frequency: Hertz, rcc: &Rcc) {
         self.stop();
-        let (psc, arr) = get_clock_config(frequency.0, self.instance.clock_frequency(rcc));
+        let (psc, arr) = get_clock_config(frequency.0, I::clock_frequency(rcc));
         self.instance.psc.write(|w| w.psc().bits(psc));
         self.instance.arr.write(|w| w.arr().bits(arr));
         self.start();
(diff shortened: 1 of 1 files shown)


azerupi:
The clippy failure is the same as what I got in #173, it has been fixed there.

hannobraun (approved):
Thank you, @azerupi!

azerupi:
Great, thanks!

Just to make sure, you did see this part?
>In the implementation I changed clock_frequency on the Instance to not take &self.

Because I think that is a breaking change since `Instance` is public. :slightly_smiling_face:

dbrgn:
Breaking changes are fine, we have breaking changes all the time. However, maybe you could send a PR that updates the CHANGELOG?

hannobraun:
I did see that, but thanks for pointing it out again. I could have been a bit less terse :slightly_smiling_face: 

While it is strictly speaking a breaking change, I don't think there's much reason for end users to touch `Instance` except in trait bounds. Its methods are more of an internal API, although Rust won't let us express that. So I think this change has pretty low impact, and given that, I believe that using `unsafe` to conjure up an `Rcc` is not worth it in this case.

Also, what @dbrgn said, we break things all the time :smile: In a perfect world, we'd take more steps to avoid/alleviate that, but we have to work with the resources we've got (on the maintainer and contributor side).

PR #176, opened 2021-06-15
Update dependencies

hannobraun:
Tested by running [this USB example](https://github.com/braun-embedded/rust-catena-4610/blob/main/examples/usb_echo.rs).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ targets = ["thumbv6m-none-eabi"]
 
 [dependencies]
 stm32l0 = "0.13.0"
-as-slice = "0.1.0"
+as-slice = "0.2.1"
 embedded-hal = { version = "0.2.3", features = ["unproven"] }
 cortex-m =  "0.7.0"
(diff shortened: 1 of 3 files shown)


hannobraun:
@stm32-rs/stm32l0  Anyone opposed to this change? This should be fairly low-impact, except for potentially the `stm32-usbd` upgrade, and I've tested verified that USB still works.

dbrgn (approved):
Looks good to me!

To be honest, I don't find the multi-line dependency declarations more readable, on the contrary, I prefer one-dependency-per-line. But I'm perfectly OK with the change 🙂

hannobraun:
> To be honest, I don't find the multi-line dependency declarations more readable, on the contrary, I prefer one-dependency-per-line. But I'm perfectly OK with the change :slightly_smiling_face:

For the record, I don't feel strongly about this issue. I'm merging this now, since you already approved and said it's okay, but I'm not opposed to someone changing that back in the future.

PR #178, opened 2021-06-23
Re-generate Cargo mcu-* features

dbrgn:
This uses the latest STM32CubeMX database.

(After merging this, the changes from #173 should only cover the new features, not add any new MCUs.)

CC @azerupi


dbrgn:
Clippy isn't happy, but that's unrelated.

Since this is purely generated code, and since it only adds more aliases without the potential of breaking anything, I'll go ahead and merge this PR without review!

PR #182, opened 2021-07-15
Fix README typo

abonnaudet-ledger:


dbrgn (approved):
Thanks for sending a fix!

PR #183, opened 2021-07-25
Migrate crate over to embedded-time

jamwaffles:


src/delay.rs
```diff
@@ -31,11 +33,17 @@ impl Delay {
         let ticks_per_us = freq / 1_000_000_u32;
         Delay { ticks_per_us, syst }
     }
+
+    /// Wait for the given time.
+    ///
+    /// Note that durations above `u32::MAX` microseconds will be clamped at `u32::MAX`.
```
jamwaffles, line 39: I'm not particularly happy with this behaviour but I can't think of a better solution right now. Thoughts appreciated!

src/delay.rs
```diff
@@ -31,11 +33,17 @@ impl Delay {
...
     pub fn delay<T>(&mut self, delay: T)
     where
-        T: Into<MicroSeconds>,
+        T: TryInto<Microseconds>,
```
jamwaffles, line 42: This is so we can use e.g. `1.second()` or `300.milliseconds()` in calling code. We could also use this pattern in other hertz-accepting places. There's at least one demo that replaces `1.khz()` with `100_000.Hz()` which I don't like.

examples/pwm.rs
```diff
@@ -24,7 +24,7 @@ fn main() -> ! {
     let gpioa = dp.GPIOA.split(&mut rcc);
 
     // Initialize TIM2 for PWM
-    let pwm = pwm::Timer::new(dp.TIM2, 10.khz(), &mut rcc);
+    let pwm = pwm::Timer::new(dp.TIM2, 10_000.Hz(), &mut rcc);
```
jamwaffles, line 27: This (and IIRC the SPI interface require this change unless I use `TryInto` in the APIs - embedded-time can convert from e.g. `Khz<u32>` to `Hz<u64>` without failing, but needs the fallible conversions if we want to go from `Khz<u32>` to `Hz<u32>`. Realistically, confining to `u32` everywhere isn't an issue (IMO) so I'd be up for using `TryInto` and clamping to `u32::MAX` like I do with the `delay()` API. Thoughts?

src/delay.rs
hannobraun: I'm not happy about that either, but can't think of a better solution right now. A fallible interface isn't great, panicking isn't great either. Your solution here at least should work for the common case, and the pitfall is documented.

src/serial.rs
```diff
@@ -93,14 +91,14 @@ pub enum StopBits {
 }
 
 pub struct Config {
-    pub baudrate: Bps,
+    pub baudrate: BitsPerSecond,
```
hannobraun, line 94: This isn't right, is it? I think we need `BytesPerSecond`.

(I don't think it makes a difference in functionality, but it's the wrong type semantically. Could lead to confusion.)

hannobraun (changes requested):
Thank you, @jamwaffles! I found one issue (see comment), but otherwise this looks great!

src/serial.rs
jamwaffles: Turns out there's a `Baud` struct which is far better suited lol. I force pushed to the branch with a fix for `BitsPerSecond` to `Baud`.

hannobraun (approved):
Thank you, @jamwaffles!

PR #184, opened 2021-07-25
Use probe-run instead of GDB for running examples

jamwaffles:
As per the title, with README instructions on how to run examples.


dbrgn (approved):
No objections from me. What about @astro or @hannobraun?

hannobraun (approved):
I like it! Thank you, @jamwaffles!

PR #185, opened 2021-07-30
Update MCU alias in README

dbrgn:
It hasn't been updated when we introduced the EEPROM and RAM/Flash features.


dbrgn:
Merging right away, since this is a trivial README change without any change in functionality.

PR #186, opened 2021-08-01
Fix max frequency assertion

jamwaffles:
IIRC I did test the examples when working on #183, but this error shows
up when using the PLL which none of them do. I found this when testing
with my own project code. At any rate, here's a fix for embedded-time
comparisons when using the PLL.


hannobraun (approved):
Thank you, @jamwaffles, good catch!

Issue #189, opened 2021-09-19
GPIO Pin#downgrade doesn't fully erase Port identifier


blakesmith:
Hi friends!

If you need to gather an array of GPIO pins, that have different types (including different port numbers), the suggested approach is to use something like `PA0.downgrade()` to erase the underlying pin identifier and port from the type.

When using the `stm32f0xx-hal` crate, [this fully erases the pin number and port from the type](https://docs.rs/stm32f0xx-hal/0.17.1/stm32f0xx_hal/gpio/gpioa/struct.PA0.html#method.downgrade). Notice how the struct returned is just a `Pin` struct, allowing pins with different ports and pin numbers to be used in an array.

The `stm32l0xx-hal` crate also implements `downgrade`, [but it only erases the pin number, and not the port](https://docs.rs/stm32l0xx-hal/0.7.0/stm32l0xx_hal/gpio/gpioa/struct.PA0.html#method.downgrade). Because it doesn't erase the port identifier, it prevents storing an array of pins with different types. Storing a separate array for each port identifier type would eliminate a single ordering across all pins.

I'm considering two ways to fix this:

1. Change `downgrade` to return a fully erased `Pin` type, instead of a port specific one. This would be a breaking change for anyone who's depending on `downgrade` to return port specific types.
2. Add a new `downgrade_fully` method, or something similarly named to return the fully erased `Pin`. Upside is: No breaking API changes with outstanding users who are dependent on the types being stable, downside is: the `downgrade_fully` would deviate from other `stm32-hal` implementations.

I can make choice 1 work for a private fork of stm32l0xx-hal, but would ideally love to be able to push this upstream so I don't need to run on a private fork anymore. I'd love to get people's weigh-in on this.

Thanks so much for the crate!

dbrgn:
I'd tend towards number 1, but are there any use cases where an erased pin but a non-erased port would be important?

jamwaffles:
My vote goes for number 1 as well

blakesmith:
Thanks for the weigh-in everyone! I'll work on a PR to change `downgrade` to remove port typing.

```
are there any use cases where an erased pin but a non-erased port would be important?
```

[...]

PR #190, opened 2021-09-20
PXi.downgrade() erases the port identifier as well

blakesmith:
PXi.downgrade() should erase the port identifier, allowing pins to be collected into an array. See full description here: https://github.com/stm32-rs/stm32l0xx-hal/issues/189.

Most of this is borrowed from [stm32f0xx-hal](https://github.com/stm32-rs/stm32f0xx-hal).

Still waiting on my board to arrive, so I can fully test this during board bring-up, but wanted to get the code up early.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -1,5 +1,6 @@
 //! General Purpose Input / Output
 
+use core::convert::Infallible;
 use core::marker::PhantomData;
 
 use crate::rcc::Rcc;
@@ -11,6 +12,13 @@ pub trait GpioExt {
 
     /// Splits the GPIO block into independent pins and registers
     fn split(self, rcc: &mut Rcc) -> Self::Parts;
+}
+
+trait GpioRegExt {
+    fn is_low(&self, pos: u8) -> bool;
+    fn is_set_low(&self, pos: u8) -> bool;
+    fn set_high(&self, pos: u8);
+    fn set_low(&self, pos: u8);
 }
 
 /// Input mode (type state)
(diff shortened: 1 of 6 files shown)


hannobraun (changes requested):
Thank you, @blakesmith! The changes look good to me, but CI is failing, which needs to be fixed.

I'd also like to hold off on merging until you (or someone else) had a chance to test this.

blakesmith:
Thanks for taking a look @hannobraun! I'll dig into the CI and try to get that cleared up.

blakesmith:
Alright. I was able to silence the warnings (that were already present) in [this commit](https://github.com/stm32-rs/stm32l0xx-hal/pull/190/commits/97c3e2f75b3ae255cc6184c19b70a0d2f2621623).

I also added an example `blink_downgrade` [in this commit](https://github.com/stm32-rs/stm32l0xx-hal/pull/190/commits/ab3be53d109db0f7a4c033faef0570ebdfacab7d), that I was able to verify is working on a custom board that uses an STM32L053R8Tx:

```
cargo objcopy --release --features mcu-STM32L053R8Tx --example blinky_downgrade -- -O ihex blinky_downgrade.hex
```

@hannobraun Let me know what you think, and if CI is looking better now!

hannobraun:
Thank you, @blakesmith. I just triggered another CI run, let's see what happens.

blakesmith:
@hannobraun I updated the [breaking examples in this commit](https://github.com/stm32-rs/stm32l0xx-hal/pull/190/commits/f20394425abf82142d9d2ce1099eb6a5617f15e3), and ran the same command as CI to verfy:

```
cross build --release --examples --features rt,mcu-STM32L011D3Px
```

All the examples are now building for me locally. That PR pretty clearly shows how code in the field that tries to update to the latest stm32l0xx-hal version would need to be modified with a similar change, if it's using the existing port-specific downgraded pin types.

Let me know what you think!

src/gpio.rs
```diff
@@ -310,17 +310,17 @@ macro_rules! gpio {
                     fn mode<M: PinMode>(&mut self) {
                         let offset = 2 * $i;
                         unsafe {
-                            &(*$GPIOX::ptr()).pupdr.modify(|r, w| {
+                            let _ = &(*$GPIOX::ptr()).pupdr.modify(|r, w| {
```
hannobraun, line 313: Hmm, can't the `&` just be removed, instead of using the borrow? Same for the other places where the warning occurred.

Just pondering. Feel free to look into it, if you want to, but this is good enough. It can be cleaned up some other time.

hannobraun (approved):
Looks good to me! CI is happy too.

So the only thing we are waiting for is for your hardware to arrive, so you can give it a quick test. Someone else jumping in and testing would be fine too, of course.

Thank you again, @blakesmith! Will merge as soon as someone confirms this works on real hardware.

src/gpio.rs
blakesmith: Good thinking, I like that way better. Will update!

blakesmith:
Thanks @hannobraun!

> So the only thing we are waiting for is for your hardware to arrive, so you can give it a quick test. Someone else jumping in and testing would be fine too, of course.

I actually was able to use my revision 1 prototype board that has a `STM32L053R8Tx` on it to verify things work using the new `blinky_downgrade` example I added.

Built using:
```
cargo objcopy --release --features mcu-STM32L053R8Tx --example blinky_downgrade -- -O ihex blinky_downgrade.hex
```

Happy to do some more testing / verification with my board where you think it's warranted!

hannobraun:
Thank you for the latest changes, @blakesmith!

> Happy to do some more testing / verification with my board where you think it's warranted!

I think this is good enough, thanks!

blakesmith:
Thank you @hannobraun for the review, and help with everything. Appreciate your effort on this board crate, it's much appreciated!

PR #191, opened 2021-10-12
RTC API revamp based on the rtcc crate

dbrgn:
The `rtc::RTC` struct has been renamed to `rtc::Rtc` and includes a big
refactoring and a few API changes. It now implements the traits from the
[rtcc crate](https://docs.rs/rtcc/) and uses date/time types from Chrono.

Because additional dependencies were added, the `rtc` module is now
behind an optional `rtc` feature (enabled by default).

The changes also include a fix for RTC year handling. Previously, the
implementation incorrectly assumed that the BCD year 00 corresponds to
1970, but this results in a wrong leap year calculation. The correct
time base is the year 2000.

### DIFF
--- a/src/rtc.rs
+++ b/src/rtc.rs
@@ -89,8 +89,8 @@ impl Rtc {
     ///
     /// # Errors
     ///
-    /// Returns [`Error::InvalidInputData`](Error::InvalidInputData) if the
-    /// `init` datetime is outside of the valid range (years 2000-2099).
+    /// Returns [`Error::InvalidInputData`] if the `init` datetime is outside
+    /// of the valid range (years 2000-2099).
     ///
     /// # Panics
     ///
@@ -169,8 +169,7 @@ impl Rtc {
     ///
     /// Note: Only dates in the range `2001-01-01 00:00:00` to
     /// `2099-12-31 23:59:59` are supported. If a date outside this range is
-    /// passed in, [`Error::InvalidInputData`](Error::InvalidInputData) will be
-    /// returned.
+    /// passed in, [`Error::InvalidInputData`] will be returned.
     pub fn set(&mut self, instant: NaiveDateTime) -> Result<(), Error> {
         // Validate and encode datetime
         let y: i32 = instant.year();


Cargo.toml
```diff
@@ -64,6 +65,12 @@ usb-device = "0.2.3"
 usbd-serial = "0.1.0"
 
 [features]
+# Default features
+default = ["rtc"]
+
+# Support for RTC, pulls in rtcc and chrono dependencies
+rtc = ["rtcc"]
```
hannobraun, line 72: Is the added complexity of having another Cargo feature worth the flexibility it provides? I'm not sure, and personally I would probably default to making the RTC and the `rtcc` dependency non-optional, unless that causes problems in practice.

But I think it's fine, definitely not a blocker.

src/rtc.rs
```diff
@@ -12,25 +36,73 @@ use crate::{
...
+    /// already configured. If the clock is not yet configured, and init is set
+    /// to `None`, then the datetime corresponding to `2000-01-01 00:00:00`
+    /// will be used for initialization.
     ///
-    /// The `initial_instant` argument will only be used, if the real-time clock
-    /// is not already configured. Its `subsecond` field will be ignore in any
-    /// case.
+    /// # Errors
+    ///
+    /// Returns [`Error::InvalidInputData`](Error::InvalidInputData) if the
```
hannobraun, line 92:
```suggestion
    /// Returns [`Error::InvalidInputData`] if the
```

That should also work, and is a bit simpler.

src/rtc.rs
```diff
@@ -89,12 +162,30 @@ impl RTC {
...
     }
 
-    /// Sets the date/time
-    pub fn set(&mut self, instant: Instant) {
-        self.write(|rtc| {
+    /// Sets the date/time.
+    ///
+    /// Note: Only dates in the range `2001-01-01 00:00:00` to
+    /// `2099-12-31 23:59:59` are supported. If a date outside this range is
+    /// passed in, [`Error::InvalidInputData`](Error::InvalidInputData) will be
```
hannobraun, line 172:
```suggestion
    /// passed in, [`Error::InvalidInputData`] will be
```

Cargo.toml
dbrgn: I don't think that's a lot of added complexity, since it's a single `#[cfg(...)]`. On the other hand, it enables people to reduce the number of dependencies they use, in case RTC is not used in a firmware project.

(Regarding binary size it probably doesn't make a relevant difference in case the functionality from RTC isn't actually used, right?)

PR #195, opened 2021-11-04
Release version 0.8.0

### LINKED ISSUE #194: New release (0.8.0)
The last release was in March, since then [quite a few things were changed](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/CHANGELOG.md). How about a new release? @hannobraun I think you tagged the last few releases, right?

hannobraun:
Close #194


hannobraun:
New version released: https://crates.io/crates/stm32l0xx-hal/0.8.0

PR #196, opened 2021-11-04
unify rcc enable/reset

burrbull:


burrbull:
Some peripheral in this hal reset before enable. This is strange.

src/pwm.rs
```diff
@@ -44,7 +44,8 @@ where
     pub fn new(timer: I, frequency: impl Into<Hertz>, rcc: &mut Rcc) -> Self {
         let frequency = frequency.into();
 
-        timer.enable(rcc);
+        I::enable(rcc);
+        I::reset(rcc);
```
hannobraun, line 48: Note: At first glance, this looks like a reset was added, but the reset was already present in the previous `enable` method. Looks correct to me.

src/timer.rs
```diff
@@ -82,18 +82,17 @@ impl TimerExt<SYST> for SYST {
...
     /// Selects the master mode for this timer
     fn select_master_mode(&mut self, variant: Self::MasterMode);
 }
 
 impl<T: GeneralPurposeTimer> Timer<T> {
-    pub fn new(mut tim: T, rcc: &mut Rcc) -> Self {
-        tim.enable(rcc);
+    pub fn new(tim: T, rcc: &mut Rcc) -> Self {
+        T::enable(rcc);
+        T::reset(rcc);
```
hannobraun, line 95: Note: At first glance, this looks like a reset was added, but the reset was already present in the previous `enable` method. Looks correct to me.

hannobraun (approved):
Thank you, @burrbull, this looks great!

How comfortable are you with the testing of this pull request? Is this ready to merge, or would you like to see more testing?

> Some peripheral in this hal reset before enable. This is strange.

I might have done that, although I've seen that before in other code (not sure whether in this HAL or elsewhere).

That's because this HAL (or any of the other STM32 HALs, as far as I'm aware) do nothing to prevent the user from changing the peripheral via the PAC API, but then expect the peripheral to be in a specific state on HAL API initialization (sometimes encoding that expectation as type state).

By doing a reset, we make sure that the expectations of the API match the actual hardware state.

burrbull:
> By doing a reset, we make sure that the expectations of the API match the actual hardware state.

I meant that usually we do enable and then reset. But I saw several times in this particular HAL reset preceded enable. For example on AES or DMA.

hannobraun:
> I meant that usually we do enable and then reset. But I saw several times in this particular HAL reset preceded enable (reset disabled peripheral). For example on AES or DMA.

Ah, got it. I would have thought that works, but I could be wrong. It might just be a bug.

Cargo.toml
```diff
@@ -54,11 +54,13 @@ rtc = ["rtcc"]
 rt = ["stm32l0/rt"]
 disable-linker-script = []
 
+device-selected = []
```
dbrgn, line 57: Maybe this should be renamed to `family-selected`? The three `stm32l0x[123]` features shouldn't really be used beyond PAC feature preselection, because they don't really say much about the features of that family. See https://github.com/stm32-rs/stm32l0xx-hal#mcu-features-mcu-.

An alternative would be to get rid of those features entirely, and to depend on `stm32l0/xxx` directly from the generated `mcu-*` features.

dbrgn (commented):
Would you mind making your commit messages a bit more descriptive? A commit message like "rcc enable" or "dsb" doesn't really tell me anything about the change.

This doesn't really change anything about the public API, right? Otherwise it would be great if you could add some upgrade notes to the `CHANGELOG.md`.

Cargo.toml
hannobraun: > An alternative would be to get rid of those features entirely, and to depend on `stm32l0/xxx` directly from the generated `mcu-*` features. What do you think, @hannobraun? (Would be out of scope for this PR though.)

I don't know whether those are used somewhere else, but sure, if it's possible, I'm all for it.

burrbull:
cc @hannobraun

PR #197, opened 2021-12-05
Setup hsi16diven when running Rcc::freeze()

airelil:
As per CubeMX, some system clock frequencies like
3Mhz, 6MHz, 8MHz or 24MHz, can be achieved only
if HSI16 divider is enabled.

RCC_CR, Bit 3 HSI16DIVEN HSI16 divider enable bit
This bit is set and reset by software to enable/disable
the 16 MHz HSI divider by 4. It can be written anytime.
0: no 16 MHz HSI division requested
1: 16 MHz HSI division by 4 requested

Example of 8MHz configuration:
```Rust
let clock_config = hal::rcc::Config::pll(
    hal::rcc::PLLSource::HSI16(true),
    hal::rcc::PLLMul::Mul4,
    hal::rcc::PLLDiv::Div2,
    );
```

### DIFF
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -11,7 +11,7 @@ use stm32l0xx_hal::{pac, prelude::*, rcc::Config};
 fn main() -> ! {
     let dp = pac::Peripherals::take().unwrap();
 
-    let mut rcc = dp.RCC.freeze(Config::hsi16(false));
+    let mut rcc = dp.RCC.freeze(Config::hsi16());
     let mut adc = dp.ADC.constrain(&mut rcc);
     let gpioa = dp.GPIOA.split(&mut rcc);
     let gpiob = dp.GPIOB.split(&mut rcc);
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -11,7 +11,7 @@ use stm32l0xx_hal::{pac, prelude::*, rcc::Config};
 fn main() -> ! {
     let dp = pac::Peripherals::take().unwrap();
 
-    let mut rcc = dp.RCC.freeze(Config::hsi16());
+    let mut rcc = dp.RCC.freeze(Config::hsi16(false));
     let mut adc = dp.ADC.constrain(&mut rcc);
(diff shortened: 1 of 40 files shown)


jamwaffles:
IMO it would be much less confusing to add a new variant, e.g. `PLLSource::HSI16Div4` instead of an ambiguous (and undocumented) `bool` - I stumbled upon its usage as a variable named `div4` only by chance.

Ideally, is it possible to calculate whether `div4` needs to be enabled or not automatically? I haven't looked at the code too closely so this might not be possible.

hannobraun (changes requested):
Thank you for the pull request, @airelil! And thank you @jamwaffles, for taking a look.

I agree with @jamwaffles that the undocumented `bool` flag is not great. Although I don't have a strong opinion on whether a struct variant (like `HSI16 { div_4: bool }`) or a separate `HSI16Div4` variant would be better.

If the need for `div_4` can be calculated automatically, all the better! But if it's not straight-forward to do, I'd be happy to merge this pull request with the explicit flag.

hannobraun:
The CI failure looks unrelated to this pull request. I've opened #199.

airelil:
That's a fair point about documenting. I'll probably change it to something like this:
```Rust
/// HSI16 divider
#[derive(Clone, Copy)]
pub enum HSI16Div {
    Div1 = 1,
    Div4 = 4,
}

/// System clock mux source
#[derive(Clone, Copy)]
pub enum ClockSrc {
    MSI(MSIRange),
    PLL(PLLSource, PLLMul, PLLDiv),
    HSE(Hertz),
    HSI16(HSI16Div),
}
```
That would be consistent with the rest of clock sources definitions.

hannobraun:
That looks good, @airelil!

airelil:
Added `HSIDiv` enum. Also reverted the changes in examples so the fix is less intrusive now. I rebased the changes to include @jamwaffles' fix for Clippy.

hannobraun (approved):
Thank you, @airelil, looks good!

PR #203, opened 2021-12-07
Fix Cargo features used by docs.rs

dbrgn:
The docs.rs build currently fails: https://docs.rs/crate/stm32l0xx-hal/0.8.0/builds/458082 Some of the required features for building were missing.

To avoid this in the future, specified an MCU feature for a random (read: the one I'm using) concrete MCU instead.

Can we re-trigger the building of the docs without creating a new release? Probably not, right? Should we create a 0.8.1 release to fix the docs? (I could do that.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ repository = "https://github.com/stm32-rs/stm32l0xx-hal"
 version = "0.8.0"
 
 [package.metadata.docs.rs]
-features = ["stm32l0x2", "rt", "stm32-usbd", "eeprom-2048", "rtc"]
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you for the pull request, @dbrgn! Looks good.

I think we'll have to put out a new release to re-trigger the docs.rs build. I think that's reasonable. We just  have to be careful about the version. Not sure, if the recent changes were non-breaking, so we might have to release 0.9.0, or backport this change to a 0.8.x release branch and release 0.8.1.

I'm fine with either option.

PR #205, opened 2021-12-09
Release 0.8.1

dbrgn:
This also adds a `RELEASING.md` file as internal documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,25 @@ forget to update the links at the bottom of the changelog as well.-->
 ### Documentation
 
 
+## [v0.8.1] - 2021-12-09
+
+### Additions
+
+- Unify RCC enable / reset ([#196])
+
+### Breaking Changes
+
+- Setup hsi16diven when running Rcc::freeze ([#197])
+
+### Non-Breaking Changes
+
+- Fix clippy lints ([#200])
+
+### Documentation
+
+- Fix Cargo features used by docs.rs ([#203])
+
+
 ## [v0.8.0] - 2021-11-04
 
 ### Additions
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,25 @@ forget to update the links at the bottom of the changelog as well.-->
...
 
+## [v0.8.1] - 2021-12-09
+
+### Additions
+
+- Unify RCC enable / reset ([#196])
+
+### Breaking Changes
+
+- Setup hsi16diven when running Rcc::freeze ([#197])
```
dbrgn, line 35: Breaking because it changes public API.

CHANGELOG.md
```diff
@@ -24,6 +24,25 @@ forget to update the links at the bottom of the changelog as well.-->
 ### Documentation
 
 
+## [v0.8.1] - 2021-12-09
+
+### Additions
+
+- Unify RCC enable / reset ([#196])
```
dbrgn, line 31: Not sure if this is an addition or breaking / non-breaking change.

dbrgn:
Note: Some changes are technically breaking changes, although it's highly unlikely that people are going to be affected. According to SemVer, below 0.x everything goes, so this would be acceptable from a spec point of view.

However, I could also change the version number to 0.9.0 instead.

jamwaffles:
IMO I think it's ok to release as 0.8.1 - this crate gets few downloads on crates.io. If upgrading to the latest `stm32l0xx` crate (waiting on https://github.com/stm32-rs/stm32-rs/pull/659 to be released) is also a breaking change, that might be a good point to release 0.9.0?

hannobraun:
> According to SemVer, below 0.x everything goes, so this would be acceptable from a spec point of view.

I think what's more relevant to us, is how Cargo *implements* SemVer. And Cargo treats 0.8.1 as compatible to 0.8.0.

I don't see a reason to cut corners here. Just call it 0.9.0. If we want to release 0.10.0 in 2 weeks, so be it.

dbrgn:
I'm sure people will have lockfiles in their firmware projects. Nevertheless, it doesn't hurt to bump to 0.9, we could also release 0.137 next week and it would still be OK 🙂

I'll create a replacement PR.

PR #206, opened 2021-12-09
Add methods to check idle, tx empty and rx not empty for serial

azerupi:
There is currently already a method `pending_event` that allows to get events when the interrupts are enabled. But the `idle` flag and the `txe` and `rxne` flags can also be used when interrupts are disabled (`idleie=0`, `txeie=0` and `rxneie=0` respectively).

This PR adds methods on `Rx` and `Tx` to read the state of those flags and reset the `idle` flag. This is inspired by https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/serial.rs#L303-L324

These methods are also more intuitive IMHO than the `pending_event` method.

We tested `is_idle` and `clear_idle_interrupt` and confirmed that it worked.
What do you guys think?

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -496,6 +496,28 @@ macro_rules! usart {
             }
 
             impl Rx<$USARTX> {
+                /// Returns true if the line idle status is set
+                /// This reads the ISR register's IDLE bit. This bit is set by hardware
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @azerupi, looks good to me!

PR #207, opened 2021-12-09
Release 0.9.0

dbrgn:
Replaces #205.

This also adds a `RELEASING.md` file as internal documentation.

PR #208, opened 2021-12-09
Make cortex-m-semihosting a dev dependency

dbrgn:
It's not actually used anywhere except in examples.

(Would be nice if we could get this into the 0.9 release #207 as well.)


hannobraun (approved):
Thank you, @dbrgn! Looks good to me!

hannobraun:
Would be even better, if we left semihosting behind completely, but oh well :grin:

PR #209, opened 2021-12-22
Fixed missing possible pins for SPI1 for STM32L0x1 family

allexoll:
This was missing other possible pins for SPI1. there still is some other to be added but i think we should find a better way to have that, kind of like it's done for serial impl...


hannobraun (approved):
Thank you, @allexoll!

I haven't checked this against the reference manual, but if it works, I'm happy.

PR #210, opened 2022-01-08
Enable USART DMA for STM32L0x1 family

valentink-eroad:
Tested the change on my STM32L051T6 board with serial_dma_async example.

I also had to fix the serial_dma_async itself as the heapless `Queue::new()` creates a queue with capacity `N-1`: https://docs.rs/heapless/0.7.1/heapless/spsc/struct.Queue.html#method.new

### DIFF
--- a/examples/serial_dma_async.rs
+++ b/examples/serial_dma_async.rs
@@ -55,9 +55,11 @@ fn main() -> ! {
         .unwrap()
         .split();
 
-    // we only have two elements for each queue, so U2 is fine (size is max 2)
-    let mut rx_buffers: Queue<Pin<DmaBuffer>, 2> = Queue::new();
-    let mut tx_buffers: Queue<Pin<DmaBuffer>, 2> = Queue::new();
+    // we only have two elements for each queue, so size is max 2 is fine.
+    // Note that due to current const generics limitations the queue capacity
+    // is calculated by heapless as N-1, so we declare it as '3-1' elements.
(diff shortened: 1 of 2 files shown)


valentink-eroad:
Sorry, missed to switch to my personal Git account, so the PR is created from my 'work' profile

examples/serial_dma_async.rs
```diff
@@ -56,8 +56,8 @@ fn main() -> ! {
         .split();
 
     // we only have two elements for each queue, so U2 is fine (size is max 2)
-    let mut rx_buffers: Queue<Pin<DmaBuffer>, 2> = Queue::new();
-    let mut tx_buffers: Queue<Pin<DmaBuffer>, 2> = Queue::new();
+    let mut rx_buffers: Queue<Pin<DmaBuffer>, 3> = Queue::new();
+    let mut tx_buffers: Queue<Pin<DmaBuffer>, 3> = Queue::new();
```
dbrgn, line 60: Why does this need to be changed? Only 2 buffers are enqueued below, right?

(If it needs to be changed, then the comment above needs to be updated as well.)

dbrgn (changes requested):

examples/serial_dma_async.rs
valentink-eroad: Thanks. Addressed in 11ddd02

dbrgn (approved):
Haven't tested this myself, but I trust your tests 🙂

PR #212, opened 2022-01-13
README: Add call for maintainers

### LINKED ISSUE #211: Stepping down as a maintainer
Hey folks!

I'd like to announce that I'm stepping down as a maintainer of this crate. [I'm off to new adventures](https://github.com/sponsors/hannobraun) and would like to be able to focus on this new direction going forward.

Since that means that @dbrgn is the only remaining active maintainer, this might be an opportunity to look for new maintainers. But that is not my choice to make. I'm will be removing myself from @stm32-rs/stm32l0 shortly after posting this.

cc @arkorobotics, @dbrgn (the other members of @stm32-rs/stm32l0)
cc @adamgreig (chief of stm32-rs)

(edit: sorry, I tagged the wrong Arko originally)

dbrgn:
I guess that should do. What do you think, @jamwaffles?

Closes #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,11 @@ stm32l0xx-hal
 
 [![Build Status](https://travis-ci.com/stm32-rs/stm32l0xx-hal.svg?branch=master)](https://travis-ci.com/stm32-rs/stm32l0xx-hal)
 
+> :warning: **Note:** The stm32l0xx HAL team is looking for more maintainers!
+> Do you think you could help out with reviewing contributions and improving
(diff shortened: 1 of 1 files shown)


dbrgn:
By the way, so far our workflow was:

- External contributions can be approved and merged by one of the maintainers
- Contributions by maintainers can be reviewed/approved by another maintainer, but are merged by the submitter

Do you think that makes sense?

jamwaffles (approved):
PR looks good!

> Do you think that makes sense?

Yep, that's how I like to work as well :)

PR #217, opened 2022-08-25
Make Exti::TriggerEdge enum behave as constant (Copy+Clone+Eq+..)

MarSik:
Exti::TriggerEdge is a plain enum and it should be possible to pass it by value without move. All the other enums in the exti module behave this way too.


jamwaffles (approved):
Looks good, thanks! Apologies for the late review.

PR #220, opened 2022-10-15
Added pause & resume to timer control

allexoll:
Small PR to add pause/resume to the timers

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -139,19 +139,28 @@ macro_rules! timers {
                 }
 
                 /// Releases the TIM peripheral
-                pub fn release(self) -> $TIM {
-                    self.tim.cr1.modify(|_, w| w.cen().clear_bit());
+                pub fn release(mut self) -> $TIM {
+                    self.pause();
                     self.tim
+                }
+
+                /// Pause counting
+                pub fn pause(&mut self) {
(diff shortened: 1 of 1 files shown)


src/timer.rs
```diff
@@ -283,6 +292,18 @@ macro_rules! linked_timers {
...
+                /// Pause counting
+                fn pause(&mut self) {
+                    self.tim_primary.cr1.modify(|_, w| w.cen().clear_bit());
+                    self.tim_secondary.cr1.modify(|_, w| w.cen().clear_bit());
+                }
+
+                /// Resume counting
+                fn resume(&mut self) {
+                    self.tim_primary.cr1.modify(|_, w| w.cen().set_bit());
+                    self.tim_secondary.cr1.modify(|_, w| w.cen().set_bit());
```
jamwaffles, line 305: The original code resumes the secondary timer first to prevent race conditions. Looks like something got switched in the refactor.

```suggestion
                    self.tim_secondary.cr1.modify(|_, w| w.cen().set_bit());
                    self.tim_primary.cr1.modify(|_, w| w.cen().set_bit());
```

jamwaffles (changes requested):
Small nit, otherwise lgtm

allexoll:
make sense, thanks!

jamwaffles (approved):
Thanks very much!

PR #221, opened 2022-11-23
Enable SPI2 on subset of stm32l0x1 devices

jcard0na:
There are members of the stm32l0x1 subfamily that thave two SPIs.  In
particular the STM32l051 and the STM32l071.
    
Compile tested as:
    
    cargo build --release --example spi2 --features mcu-STM32L051C8Tx
(builds)
    
    cargo build --release --example spi2  --features mcu-STM32L031C6Tx
(fails to build, as this subfamily only has 1 SPI)
    
but not run on actual hardware.

### DIFF
--- a/examples/spi2.rs
+++ b/examples/spi2.rs
@@ -1,0 +1,55 @@
+//! An example to show the configuration and usage of SPI1 and SPI2 on `stm32l0x2`, `stm32l0x3` and
+//! the subset of `stm32l0x1` series that have two SPI ports
+//!
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_halt;
+
+use cortex_m_rt::entry;
+use stm32l0xx_hal::{pac, prelude::*, rcc::Config, spi};
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Configure the clock.
+    let mut rcc = dp.RCC.freeze(Config::hsi16());
+
+    // Acquire the GPIOA peripheral. This also enables the clock for GPIOA in
+    // the RCC register.
+    let gpioa = dp.GPIOA.split(&mut rcc);
+
+    let mut nss = gpioa.pa4.into_push_pull_output();
+    let sck = gpioa.pa5;
+    let miso = gpioa.pa6;
+    let mosi = gpioa.pa7;
+
+    // Initialise the SPI1 peripheral.
+    let mut spi = dp
+        .SPI1
+        .spi((sck, miso, mosi), spi::MODE_0, 100_000.Hz(), &mut rcc);
+
+    let gpiob = dp.GPIOB.split(&mut rcc);
+
+    let mut nss2 = gpiob.pb12.into_push_pull_output();
+    let sck2 = gpiob.pb13;
+    let miso2 = gpiob.pb14;
+    let mosi2 = gpiob.pb15;
+
+    // Initialise the SPI2 peripheral.
+    let mut spi2 = dp
+        .SPI2
+        .spi((sck2, miso2, mosi2), spi::MODE_0, 100_000.Hz(), &mut rcc);
+
+    loop {
+        nss.set_low().unwrap();
+        spi.write(&[0, 1]).unwrap();
+        nss.set_high().unwrap();
+
+        nss2.set_low().unwrap();
+        spi2.write(&[0, 1]).unwrap();
+        nss2.set_high().unwrap();
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ make upgrading the HAL as painless as possible! If it makes sense, feel free
 to add upgrade notes and examples. When adding an issue or PR reference, don't
 forget to update the links at the bottom of the changelog as well.-->
 
+- Enable SPI2 on subset of stm32l0x1 devices ([#221])
+
 ### Additions
 
 - Add `pause` and `resume` methods to timers ([#220])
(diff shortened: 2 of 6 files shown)


examples/spi2.rs
```diff
@@ -0,0 +1,52 @@
...
+        .spi((sck, miso, mosi), spi::MODE_0, 100_000.Hz(), &mut rcc);
+
+    let gpiob = dp.GPIOB.split(&mut rcc);
+
+    let mut nss2 = gpiob.pb12.into_push_pull_output();
+    let sck2 = gpiob.pb13;
+    let miso2 = gpiob.pb14;
+    let mosi2 = gpiob.pb15;
+
+    // Initialise the SPI peripheral.
```
jamwaffles, line 38:
```suggestion
    // Initialise the SPI2 peripheral.
```

examples/spi2.rs
```diff
@@ -0,0 +1,52 @@
...
+    // Acquire the GPIOA peripheral. This also enables the clock for GPIOA in
+    // the RCC register.
+    let gpioa = dp.GPIOA.split(&mut rcc);
+
+    let mut nss = gpioa.pa4.into_push_pull_output();
+    let sck = gpioa.pa5;
+    let miso = gpioa.pa6;
+    let mosi = gpioa.pa7;
+
+    // Initialise the SPI peripheral.
```
jamwaffles, line 26:
```suggestion
    // Initialise the SPI1 peripheral.
```

jamwaffles (changes requested):
According to RM0377 Table 2, "Overview of features per category" (ref  Table 1 above it), the L011 and L031 don't have SPI2, so I don't think it's the right thing to remove the feature gates completely.

Thankfully it looks like the categories mentioned in that table are mirrored in [these somewhat confusingly named features](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/Cargo.toml#L75-L78) so you could try using those instead. I'd want to diff the `cargo doc` output to check for any regressions in other parts if you do that though.

jcard0na:
Thanks for the quick review.  I think I addressed all the comments except that I don't know how to review changes to `cargo doc`.  Can you help with that?

jamwaffles:
> Thanks for the quick review.

Any time!

It looks like there are some remaining `NaiveDate` errors to sort out :(. Are you able to access the Github Actions logs on this PR? Either way, you should be able to run the commands in `.github\workflows\ci.yaml` to test for issues on your dev machine.

>  I don't know how to review changes to cargo doc

I'm not sure what you mean by this I'm afraid. I see warnings when I run `cargo doc --features 'rt mcu-STM32L011D3Px'` but nothing that stops a build, unless I try `cargo doc --release --examples --features 'rt mcu-STM32L011D3Px'` which vomits over a conflicting panic dependency. Do you mean these or is there something else coming up?

jcard0na:
> It looks like there are some remaining `NaiveDate` errors to sort out :(. Are you able to access the Github Actions logs on this PR? Either way, you should be able to run the commands in `.github\workflows\ci.yaml` to test for issues on your dev machine.

Duh.  I did replace all the deprecated functions... but I introduced a new call to `and_hms()`.  Fixed now.
 
> > I don't know how to review changes to cargo doc
> 
> I'm not sure what you mean by this I'm afraid. I see warnings when I run `cargo doc --features 'rt mcu-STM32L011D3Px'` but nothing that stops a build, unless I try `cargo doc --release --examples --features 'rt mcu-STM32L011D3Px'` which vomits over a conflicting panic dependency. Do you mean these or is there something else coming up?

I was trying to address your original `cargo doc` comment: 
>> I'd want to diff the cargo doc output to check for any regressions in other parts if you do that though.

jcard0na:
I do get an error when running:
```
cargo doc --release --examples --features 'rt mcu-STM32L011D3Px'
```
but the same error show up before my change.

```
warning: build failed, waiting for other jobs to finish...
error: duplicate lang item in crate `panic_semihosting`: `panic_impl`.
  |
  = note: the lang item is first defined in crate `panic_halt` (which `timer_interrupt_rtic` depends on)
  = note: first definition in `panic_halt` loaded from /opt/javier/dev/stm32l0xx-hal/target/thumbv6m-none-eabi/release/deps/libpanic_halt-93ea9eb1b94cbf3a.rmeta
  = note: second definition in `panic_semihosting` loaded from /opt/javier/dev/stm32l0xx-hal/target/thumbv6m-none-eabi/release/deps/libpanic_semihosting-aeee2f7c58ae038e.rmeta
  ```
I can complete the build if I comment out all `extern crate panic_halt` declarations.

**Comment by jamwaffles** [comment]
Right, that makes more sense now. It should work if you remove the `--examples` arg - those don't generate any docs.

If you wanted to fix the root cause, it would be great to have every example use `panic_halt` and remove `panic_semihosting` completely. Would also be a good opportunity to replace `extern crate panic_halt` with `use panic_halt as _;` which is a bit more modern. I'm happy to hack on this at a later date if you just want to land the SPI2 stuff for now :)

**Comment by jcard0na** [comment]
I did try
> to have every example use panic_halt and remove panic_semihosting completely.

but that still failed with the error
```
error: duplicate lang item in crate `panic_semihosting`: `panic_impl`.
```
Here is the diff in case you want to check it out [panic_halt.diff.txt](https://github.com/stm32-rs/stm32l0xx-hal/files/10098338/panic_halt.diff.txt)

Given that I don't really know where this conflicting dependency comes from, I think I'll accept your offer to land the SPI2 fix now.  Thanks!

jamwaffles (approved):
> Given that I don't really know where this conflicting dependency comes from, I think I'll accept your offer to land the SPI2 fix now.

Sure thing, and sorry the fix wasn't as easy as it should be :(. I'll put a follow up PR together to clean that stuff up later. Thanks for the attempt at least!

PR looks good too :)

PR #222, opened 2023-01-09
Disable regulator when entering Stop on STM32l0x1

jcard0na:
Only category 1 devices of the STM32l0x1 family have a `lpds` bit on the power configuration register.  Categories 2 and above use the `lpsdsr` bit instead.  On STM32l0x2 and STM32l0x3, these two bits are aliased in the stm32l0 crate, so `cr.lpds()` and `cr.lpsdsr()` will modify the same bit.

Tested by compiling the `pwr.rs` example and running it on a STM32L051C8Tx. With the fix, the stop current is reduced by 13.4 uA.

cargo build --release --example pwr --features mcu-STM32L051C8Tx

The problem could not be measured in the `pwr.rs` example because in it the device is put into low-power sleep mode before entering stop. That code accidentally left the `lpsdsr` bit set which would then affect the stop mode that followed.  Fix this by properly exiting low power sleep.

### DIFF
--- a/src/pwr.rs
+++ b/src/pwr.rs
@@ -336,7 +336,13 @@ impl PowerMode for StopMode<'_> {
             // Enter Stop mode
             w.pdds().stop_mode();
             // Disable internal voltage regulator
-            w.lpds().set_bit()
+            // Cat 1 devices of stm32l0x1 family do this differently than the rest
+            if cfg!(feature = "io-STM32L021") {
+                w.lpsdsr().clear_bit();
+                w.lpds().set_bit()
+            } else {
+                w.lpsdsr().set_bit()
+            }
         });
 
         // Wait for WUF to be cleared
--- a/examples/pwr.rs
+++ b/examples/pwr.rs
@@ -70,6 +70,7 @@ fn main() -> ! {
 
     // 5 seconds of low-power sleep mode
     exti.wait_for_irq(exti_line, pwr.low_power_sleep_mode(&mut scb, &mut rcc));
+    pwr.exit_low_power_run_mode();
     timer.wait().unwrap(); // returns immediately; we just got the interrupt
 
     blink(&mut led);


src/pwr.rs
```diff
@@ -336,7 +336,13 @@ impl PowerMode for StopMode<'_> {
             // Enter Stop mode
             w.pdds().stop_mode();
             // Disable internal voltage regulator
-            w.lpds().set_bit()
+            // Cat 1 devices of stm32l0x1 family do this differently than the rest
+            if cfg!(feature = "stm32l0x1") && cfg!(feature = "io-STM32L021") {
```
jamwaffles, line 340: Shouldn't this just be:

```suggestion
            if cfg!(feature = "stm32l0x1") {
```

The features in `Cargo.toml` specify `io-STM32L021` for anything with `stm32l0x1` so I'm not sure that other feature adds much.

If you'd like to keep it, shouldn't it be

```suggestion
            if cfg!(feature = "stm32l0x1") || cfg!(feature = "io-STM32L021") {
```

instead?

jamwaffles (changes requested):
Apologies for the delay. Just a question around the features used otherwise lgtm!

I'm not sure what caused Clippy to start whinging but it doesn't look related to your changes so don't worry about it if you were :)

jamwaffles:
I just merged #223 which should fix this branch too if you rebase from `master`.

src/pwr.rs
jcard0na: I believe you meant to drop `cfg!(feature = "stm32l0x1")` as checking for `cfg!(feature = "io-STM32L021")` would be sufficient If so, the first suggested change above is incorrect and should be:
```
if cfg!(feature = "io-STM32L021") {
```
I was being extra cautious in case someone in the future might mix `"io-STM32L021"` with a non-`"stm32l0x1"` device.  That is why I had the `&&`, meaning this is a "category 1" device of the "stm32l0x1" family.  Just for reference, below is the reference manual section that documents this (see note at bottom):
 
image

I'll go ahead and change and pull #223 as well.

Thanks!

jamwaffles: That works too :) thanks for the explanation

jamwaffles (approved):

jamwaffles:
Thanks!

src/pwr.rs
dbrgn: Regarding the io features, see https://github.com/stm32-rs/stm32l0xx-hal#io-features-io- . As far as I'm aware, these features map directly to the device category (but we re-used the terminology from STM32CubeMX). Checking just the io-feature should be safe.

PR #226, opened 2023-03-26
Allow pin modes OpenDrain and PushPull for SPI output pins

osannolik:
I needed to use `OpenDrain` mode for SPI output pins, but the init for SPI were only defined for pins set to `Analog` (i.e. implicitly `PushPull`). Therefore I added `Output<OpenDrain>` as an alternative and also `Output<PushPull>` to allow for more explicit user peripheral setup code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ forget to update the links at the bottom of the changelog as well.-->
 
 ### Additions
 
+- SPI: Allow output pins of mode `PushPull` and `OpenDrain` ([#226])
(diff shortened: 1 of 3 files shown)


jamwaffles (changes requested):
Thanks for the PR! Changes look good, and according to ST AN4899 section 4.3.3:

> The output buffer can be configured in open-drain or push-pull mode

This is a supported feature so we're all good.

Can you please add a changelog entry under [the `Additions` heading](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/CHANGELOG.md#additions)?

Clippy is also whinging in CI but that looks like an easy fix. Thanks!

jamwaffles (approved):
Looks good, thanks!

PR #228, opened 2023-03-30
Skip /32 AHB prescaler

### LINKED ISSUE #225: Incorrect AHB prescaler calculation
I think the AHB prescaler divisors are not correctly calculated.
The enum kinds [here](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/rcc.rs#L80-L90) are mapped to four bits which are then used to left-shift the divisor in [here](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/rcc.rs#L419-L422).
However, there's a gap in the divisors, there is no Div32, but the bits-to-divisor calculation does not handle this gap.
This means, that when a prescaler of 64 or higher is used, the reported frequency is too high by a factor of 2.

knoellle:
Fixes #225 

There is no /32 AHB prescaler, but the old calculation code assumed there was, causing all calculated clock frequencies with an AHB prescaler > 16 to be off by a factor of 2.

Note that this only affects the frequencies stored in the `Clocks` struct, not the ones actually applied to the AHB frequency.
Page 194 of the [reference manual](https://www.st.com/resource/en/reference_manual/rm0367-ultralowpower-stm32l0x3-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) shows mapping of register bits to prescaler factors. As far as I can see, those are implemented correctly.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -413,6 +413,8 @@ impl RccExt for RCC {
 
         let ahb_freq = match cfgr.ahb_pre {
             AHBPrescaler::NotDivided => sys_clk,
+            // skip /32
+            pre if pre as u8 & 0b0100 > 0 => sys_clk / (1 << (pre as u8 - 6)),
             pre => sys_clk / (1 << (pre as u8 - 7)),
         };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,6 +190,7 @@ _Not yet tracked in this changelog._
 <!-- Links to pull requests and issues. -->
 
 [#226]: https://github.com/stm32-rs/stm32l0xx-hal/pull/226
+[#228]: https://github.com/stm32-rs/stm32l0xx-hal/pull/228
 [#224]: https://github.com/stm32-rs/stm32l0xx-hal/pull/224
 [#221]: https://github.com/stm32-rs/stm32l0xx-hal/pull/221
 [#220]: https://github.com/stm32-rs/stm32l0xx-hal/pull/218
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ forget to update the links at the bottom of the changelog as well.-->
 
 ### Fixes
 
+- Correct calculation of AHB prescaler for factors > 16 ([#228])
+
 ### Documentation
 
 ## [v0.10.0] - 2022-08-15


src/rcc.rs
```diff
@@ -418,6 +418,8 @@ impl RccExt for RCC {
 
         let ahb_freq = match cfgr.ahb_pre {
             AHBPrescaler::NotDivided => sys_clk,
+            // skip /32
+            pre if pre as u8 & 0b0100 > 0 => sys_clk / (1 << (pre as u8 - 6)),
```
jamwaffles, line 422: Won't this panic because `0b0100` - 6 will always underflow? I could be misunderstanding the logic here but it would be good to confirm this on real hardware if you haven't already.

jamwaffles (changes requested):
Thanks for the PR! I left a question inline, but on top of that could you also please add a changelog entry [under the `Fixed` section](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/CHANGELOG.md#fixes)?

src/rcc.rs
knoellle: if it was just `0b01xx`, then yes.
But since except for the `NotDivided` enum kind, all of the `u8` values assigned to the enum kinds have the fourth bit (index 3) set, subtracting 6 is fine.
If you look closely, I'm actually subtracting _less_ here than in the third match branch, which was the original implementation.

Honestly, I wasn't sure how to implement this. My preferred solution would be to simply define a function on the enum type that returns the prescaler factor as an integer using a large match statement. That would be much easier to read, understand, and check for errors such as this.
But I didn't want to mix styles too much, so I went with this implementation instead.

jamwaffles: Thanks for the explanation and the test on hardware - much appreciated! Makes sense now :)

jamwaffles:
This is good to go aside from:

> on top of that could you also please add a changelog entry [under the `Fixed` section](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/CHANGELOG.md#fixes)? 

If you missed it :)

knoellle:
Added a changelog entry.
However, at least on the [preview in my fork](https://github.com/knoellle/stm32l0xx-hal/blob/fixAHBPrescaler/CHANGELOG.md), the PR link doesn't seem to work. Is that because this PR was created after the fork? All the other links seem to correctly reference PRs in the upstream repository. Will that fix itself once the change is merged?

Also, in the (now obsolete) workflow run, the clippy check failed. Do you want me to fix that in this PR even though it's not a related change?

jamwaffles:
You need to add a line at the bottom of the file [like this](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/CHANGELOG.md?plain=1#L190), then the link should work.

jamwaffles:
> Also, in the (now obsolete) workflow run, the clippy check failed. Do you want me to fix that in this PR even though it's not a related change?

Yeah that's fine, thanks for bringing it up :)

knoellle:
I'm confused... why did the clippy check succeed this time? 👀 
Rustup reportedly installed the same versions in both runs.

knoellle:
Fixed the changelog link, feel free to merge when you're happy.

jamwaffles:
Thanks!

PR #232, opened 2024-01-02
Fix error "assigning to &T is undefined behavior" and clippy warnings

azerupi:
Since Rust 1.73.0 there seems to be [a new deny-by-default lint](
https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#invalid-reference-casting) to catch invalid casting from references to mutable pointers.

As far as I understand the compiler is making assumptions that references are read only and passing those assumptions on to LLVM. The only way to tell the compiler to not make those assumptions is to use `UnsafeCell`.

So this PR is essentially wrapping those casts with `UnsafeCell` and getting the mutable pointer through it.

In addition to that I fixed some warnings from compiler lints and clippy lints.

### DIFF
--- a/src/exti.rs
+++ b/src/exti.rs
@@ -69,7 +69,7 @@ impl Exti {
 
         unsafe {
             match line {
-                0 | 1 | 2 | 3 => {
+                0..=3 => {
                     syscfg.syscfg.exticr1.modify(|_, w| match line {
                         0 => w.exti0().bits(port_bm),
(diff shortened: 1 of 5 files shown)


jamwaffles (approved):
Looks good to me, although I'd like at least one more pair of eyes on this before merge if possible.

azerupi:
No worries, thank you very much!

